PR #2, opened 2018-02-02
Support for running with threaded Chez Scheme builds

weinholt:
Hello,

I tried to get Swish running on Debian and saw a problem. As a general principle, when Debian packages a piece of software, it is built with all features enabled. In Chez's case this means that threading is enabled. In suspect that Swish's process model is not compatible with threading, but I think it is up to the developer to not use threading in a bad way. Please consider these changes that makes Swish run on threaded Chez builds.

There is one remaining issue and that is the location of scheme.h. As it stands right now the makefiles hard code both the Chez Scheme version and machine type. So it will also break when 9.5.1 comes out. How do you feel about calling scheme from the makefiles, to get the version and the machine type? An alternative, which may be more pleasant in the long run, would be that Chez Scheme installs a pkg-config file. (Another complication is that Debian puts the header in /usr/include/<gnu-build-triple>, but I'm seriously considering changing this).

/Göran

### DIFF
--- a/doc/swish/gatekeeper.tex
+++ b/doc/swish/gatekeeper.tex
@@ -61,11 +61,11 @@ records, each of which has the following fields:\antipar
 
 \genserver{gatekeeper}{init} The gatekeeper \code{init} procedure
 hooks the system primitives listed in the introduction so that they
-use \code{with-mutex} with a timeout of one minute, and it sets the
-\code{current-expand} parameter to the hooked \code{sc-expand}
-procedure. The process traps exits so that \code{terminate} can unhook
-the system primitives when the process is shut down. It returns an
-empty list of \code{<mutex>} records.
+use \code{with-gatekeeper-mutex} with a timeout of one minute, and it
+sets the \code{current-expand} parameter to the hooked
+\code{sc-expand} procedure. The process traps exits so that
+\code{terminate} can unhook the system primitives when the process is
+shut down. It returns an empty list of \code{<mutex>} records.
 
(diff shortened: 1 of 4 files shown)


burgerrg:
Thanks for thinking about systems where `scheme` is the threaded version. Although Swish isn't thread-safe, it could be used in multi-threaded Chez Scheme. Please update your pull request to rename Swish's `with-mutex` to `with-gatekeeper-mutex` in order to avoid confusion in the future.

With regard to the location of scheme.h, I like your ideas and haven't had time yet to work on that. Please submit a pull request if you're interested.

weinholt:
I've updated the branch according to your wish.

As for Debian's location of scheme.h, it will go back to the normal location when the next Chez release is out. Which idea did you like, calling scheme to get the version and machine type, or the pkg-config one?

burgerrg:
Thanks, I merged the changes.

burgerrg:
Calling scheme to get the machine type and version number sounds straightforward. I haven't used pkg-config, but it also is a possibility, and I already have it installed on all 3 platforms Swish supports, so that would work, too.

burgerrg:
Maybe we should start by calling scheme and then look into submitting a change to Chez Scheme for a pkg-config file?

burgerrg:
The advantage of calling scheme is that it will work on Windows. Windows doesn't support pkg-config natively but through Cygwin, and we're not using Cygwin's compiler in Windows.

burgerrg:
Swish now calls scheme from the makefiles as desired. Thanks, @owaddell!

PR #3, opened 2018-02-05
Allow spaces in project path on windows

kleinpa:


burgerrg:
Thanks for this fix! In case the DOS path doesn't exist (because the system turned off 8.3 naming on an NTFS volume, for example), try this instead:

`cmd /c "$(cygpath -w "$(dirname "$0")/vs32.bat")" "$@"`

The same goes for vs64.

burgerrg:
I made an even simpler change to allow for spaces. Thank you for the pull request!

PR #4, opened 2018-02-08
Dynamic paths for Chez's include/lib dirs

weinholt:
Here are changes to the makefiles that remove the hardcoded paths. I've been able to test this with ta6le, partially with ta6nt (no VS installed), and with arm32le (albeit on an aarch64 machine).


burgerrg:
Thank you!

PR #5, opened 2018-02-09
enabled use of Anaconda Python

laqrix:
Here are changes to allow the use of Anaconda Python in Windows.


burgerrg:
Thanks for the suggestion! Another pull request did some of these changes, and I just pushed a change that allows one to override SCHEME and PYTHON.

Please update the build instructions for Anaconda Python and double-check the line that exports PYTHON. I think it should be `PYTHON="$(cygpath -w "$(type -p python)")" ./go`

laqrix:
Instructions updated.

PR #7, opened 2018-03-31
Minor changes to support Alpine linux

kleinpa:
These commits remove the bash and glibc dependencies.

There is no obvious portable linux implementation of `osi_get_bytes_used` so I've modified it to return 0 when compiled without glibc. The only other solution I found is [reading the current memory usage from /proc](https://gist.github.com/kleinpa/077a97302b3ea096991ad8e5505dc0f8) but thats slow, less accurate, and could be easily implemented in scheme. I'm not convinced `osi_get_bytes_used` should be in the osi at all.

### DIFF
--- a/doc/run-latex
+++ b/doc/run-latex
@@ -1,11 +1,11 @@
-#!/bin/bash -e
+#!/bin/sh -e
 if [ -t 1 ]
 then
   MODE=spinner
 else
   MODE=silent
 fi
-if [ "$1" == "-q" ]; then
+if [ "$1" = "-q" ]; then
   MODE=silent
   shift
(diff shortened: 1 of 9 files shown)


burgerrg:
Does the `-e` switch work on vanilla /bin/sh?

kleinpa:
Yep, it's part of [the standard](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#set) and works in [Alpine's sh](https://en.wikipedia.org/wiki/Almquist_shell).

burgerrg:
Thank you for this change!

Issue #14, opened 2018-11-16
Build error on macOS: symbol(s) not found


qzivli:
```
gcc -m64 -fPIC -Wall -Wextra -Werror -O2 -o ../../build/release/bin/swish main.o ../../build/release/lib/swish_kernel.o -liconv -lm -lncurses
Undefined symbols for architecture x86_64:
  "_libiconv", referenced from:
      _s_iconv_from_string in swish_kernel.o
      _s_iconv_to_string in swish_kernel.o
  "_libiconv_close", referenced from:
      _s_iconv_close in swish_kernel.o
  "_libiconv_open", referenced from:
      _s_iconv_open in swish_kernel.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[1]: *** [../../build/release/bin/swish] Error 1
make: *** [swish] Error 2
```

This need CPPFLAGS and LDFLAGS, e.g., `./configure CPPFLAGS="-I/usr/local/opt/libiconv/include" LDFLAGS="-L/usr/local/opt/libiconv/lib" `

PR #15, opened 2018-11-16
Added: CPPFLAGS, CFLAGS and LDFLAGS

qzivli:
This is related to #14


qzivli:
Now this PR has two changes:
1. Added flags to fix "symbol(s) not found" error;
2. Use GitHub mirror for chromium/gyp to make Chinese (and maybe others) users happy.

qzivli:
@burgerrg I agree that `CPPFLAGS` is not a good name, but usually, by convention`CPPFLAGS` means passing to the C or C++ preprocessor, and ``CXXFLAGS` means using it when calling the C++ compiler.

burgerrg:
Thank you for the clarification. CPPFLAGS is for the C pre-processor. Should we include CC also for the C compiler? Then the default make rule with $(CC) $(CPPFLAGS) $(CFLAGS) would work.

qzivli:
I'm working on test building swish on some fresh install OSes, some (e.g., Debian stretch) raise an error:
```
make -C src/swish all
error: unknown option `exclude=*'
usage: git describe [<options>] [<commit-ish>...]
   or: git describe [<options>] --dirty

    --contains            find the tag that comes after the commit
    --debug               debug search strategy on stderr
    --all                 use any ref
    --tags                use any tag, even unannotated
    --long                always use long format
    --first-parent        only follow first parent
    --abbrev[=<n>]        use <n> digits to display SHA-1s
    --exact-match         only output exact matches
    --candidates <n>      consider <n> most recent tags (default: 10)
    --match <pattern>     only consider tags matching <pattern>
    --always              show abbreviated commit object as fallback
    --dirty[=<mark>]      append <mark> on dirty working tree (default: "-dirty")

Makefile:46: recipe for target 'git.revision' failed
make[1]: *** [git.revision] Error 129
Makefile:10: recipe for target 'swish' failed
make: *** [swish] Error 2
```

Which the git version is:
```
# git --version
git version 2.11.0
```

The problem is the "--exclude" option of "git describe" was added until version 2.13.0

qzivli:
The problem of git version should be fixed or documented. I put an example here to help someone to install git manually:
```
apt-get install -y libssl-dev
apt-get install -y libgcrypt11-dev zlib1g-dev
apt-get install -y libcurl4-openssl-dev
apt-get install -y libexpat1-dev
apt-get install -y tcl tk
apt-get install -y gettext


wget https://github.com/git/git/archive/v2.19.1.tar.gz
tar zxvf v2.19.1.tar.gz 
cd git-2.19.1/
make
make install

# If the install directory (e.g., /root/bin) is not in PATH, put it into PATH
export PATH=/root/bin:$PATH

git --version
```

burgerrg:
Thanks for the note about the required git version.

burgerrg:
1. Please update configure to write the CC, CPPFLAGS, CFLAGS, and LDFLAGS definitions directly into src/swish/Makefile instead of src/swish/Mf-config.
1. Please fix the misspelling of LDFLAGS in README.md.
1. Please rebase and squash to a single commit.

Thanks!

burgerrg:
Thanks for the changes! We will incorporate them soon.

burgerrg:
I just pushed changes that incorporate your work. Please let me know if it works for you.

qzivli:
I just test it on my laptop, it works.

PR #20, opened 2019-01-10
Add more theory of operation to the operating system interface

burgerrg:


doc/swish/osi.tex
```diff
@@ -32,33 +32,42 @@
 \section {Theory of Operation}
 
 The operating system interface is written in C99~\cite{c99} as a
-shared library that is statically linked against the libuv and SQLite
-libraries and dynamically linked against single-threaded Chez Scheme
-9.5.  Please refer to Chapter~4 of the \emph{Chez Scheme Version 9
-  User's Guide}~\cite{chez-scheme-users-guide} for information on the
-foreign function interface. C++ is not used because C++ destructors
-may interact badly with setjmp/longjmp, used by Chez Scheme.
+shared library that links to the single-threaded Chez Scheme 9, libuv,
```
owaddell: Suggestion: delete "single-threaded" here since: (1) threading model is covered in the next paragraph and (2) it appears to imply single-threaded libuv and SQLite where the original text was careful to restrict that modifier to Chez Scheme.

Suggestion: delete the "9" here since we don't mention libuv and SQLite versions here and the reference to Chapter 4 of CSUG contains the version number.

So: "... links to the Chez Scheme, libuv, and SQLite libraries."

doc/swish/osi.tex
```diff
@@ -32,33 +32,42 @@
...
 The operating system interface uses port objects for files, console
 input, pipes to other processes, and TCP/IP connections. A port object
 is created by the various open functions, which return a port handle
 that is used for read, write, and close operations. Once a port is
 closed, its port object is freed.
 
+Whenever Scheme receives a handle to an object allocated in the C
+heap, the handle is wrapped in a Scheme record that is registered with
+an appropriate guardian. After each garbage collection, the finalizers
+use these guardians to free the objects from the C heap.
```
owaddell: Suggestion: Consider adding a reference here to the description of the finalizer process in the erlang.tex theory of operations.

doc/swish/osi.tex
burgerrg: Good ideas, thanks!

doc/swish/osi.tex
burgerrg: Good idea, thanks!

burgerrg:
I added a paragraph about worker threads, async handles, and a caution about the libuv thread pool.

doc/swish.sty
```diff
@@ -39,7 +39,7 @@
   {\end{argtbl}}
 \newenvironment{tupledef}[1]{\begin{tuple}\codeindex{#1}\end{tuple}\antipar
   \begin{argtbl}}{\end{argtbl}}
-\newcommand{\concern}[1]{\textbf{Concern:} #1}
-\newcommand{\mitigation}{\textbf{Mitigation: }}
```
owaddell: What do you think about \textsc as a middle ground between \textbf and nothing?

burgerrg: Small caps was pretty intense. What do you think of slant font? I've pushed it for your consideration.

owaddell: Slant font looks good to me.

PR #22, opened 2019-03-18
Windows: Compile using dynamic-link libraries

laqrix:
This pull request provides support for building additional DLLs by changing Swish to link against libuv, sqlite3, and osi dynamically.


laqrix:
The goal is to enable the creation of foreign libraries that can call Scheme, Swish, LibUV, and SQLite API functions.  We also want to be precise about the built-in library code loaded into the process.

On Windows we accomplish this using dynamic linking but control the load process by explicitly calling LoadLibrary with absolute paths. On other platforms, static linking achieves these goals.

- Move necessary exports into swish.h
- Load Chez Scheme DLL using standard Windows Search Order
- Load Swish DLLs (libuv, sqlite3, osi) only from application directory
- Add osi-bootstrap.ss to load Swish DLLs when not using Swish's executable
- Add SCHEME_LIB_PREFIX to use the configured version of Chez Scheme
- Cleanup usage of Mf-config variables
- Remove swish_kernel.obj

Thanks @burgerrg for all your help.

PR #25, opened 2019-04-09
Provide file-descriptor I/O

burgerrg:
This pull request adds open-fd-port for performing asynchronous I/O to file descriptors 0 (stdin), 1 (stdout), 2 (stderr), and higher. It also exports some record-related procedures from the base system.

### DIFF
--- a/doc/swish/db.tex
+++ b/doc/swish/db.tex
@@ -452,6 +452,33 @@ By default, \var{symbol->sql} is the identity function.
 
 \end{itemize}
 
+\defineentry{database?}
+\begin{procedure}
+  \code{(database? \var{x})}
+\end{procedure}
+\returns{} a boolean
+
+The \code{database?} procedure determines whether or not the datum
+\var{x} is a database record instance.
+
+\defineentry{database-create-time}
+\begin{procedure}
+  \code{(database-create-time \var{db})}
+\end{procedure}
+\returns{} a clock time in milliseconds
+
+The \code{database-create-time} procedure returns the clock time from
+\code{erlang:now} when database record instance \var{db} was created.
+
+\defineentry{database-filename}
+\begin{procedure}
+  \code{(database-filename \var{db})}
+\end{procedure}
+\returns{} a string
+
+The \code{database-filename} procedure returns the filename of
+database record instance \var{db}.
+
 \defineentry{database-count}
 \begin{procedure}
   \code{(database-count)}
@@ -470,6 +497,43 @@ databases.
 The \code{print-databases} procedure prints information about all open
 databases to textual output port \var{op}, which defaults to the
 current output port.
+
+\defineentry{statement?}
+\begin{procedure}
+  \code{(statement? \var{x})}
+\end{procedure}
+\returns{} a boolean
+
+The \code{statement?} procedure determines whether or not the datum
+\var{x} is a statement record instance.
+
+\defineentry{statement-create-time}
+\begin{procedure}
+  \code{(statement-create-time \var{stmt})}
+\end{procedure}
+\returns{} a clock time in milliseconds
+
+The \code{statement-create-time} procedure returns the clock time from
+\code{erlang:now} when statement record instance \var{stmt} was
+created.
+
+\defineentry{statement-database}
+\begin{procedure}
+  \code{(statement-database \var{stmt})}
+\end{procedure}
+\returns{} a string
+
+The \code{statement-database} procedure returns the database record
+instance of the statement record instance \var{stmt}.
+
+\defineentry{statement-sql}
+\begin{procedure}
+  \code{(statement-sql \var{stmt})}
+\end{procedure}
+\returns{} a string
+
+The \code{statement-sql} procedure returns the SQL string of the
+statement record instance \var{stmt}.
 
 \defineentry{statement-count}
 \begin{procedure}
--- a/doc/swish/erlang.tex
+++ b/doc/swish/erlang.tex
@@ -1348,6 +1348,17 @@ The \code{listener-address} procedure returns the \code{address} of
 the given TCP \var{listener}\index{TCP listener}.
 
 % ----------------------------------------------------------------------------
+\defineentry{listener-create-time}
+\begin{procedure}
+  \code{(listener-create-time \var{listener})}
+\end{procedure}
+\returns{} a clock time in milliseconds
+
+The \code{listener-create-time} procedure returns the clock time from
+\code{erlang:now} when the given TCP \var{listener}\index{TCP
+  listener} was created.
+
+% ----------------------------------------------------------------------------
 \defineentry{listener-port-number}
 \begin{procedure}
   \code{(listener-port-number \var{listener})}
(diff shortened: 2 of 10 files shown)


src/swish/io.ss
```diff
@@ -54,8 +58,12 @@
    open-file-to-write
    open-utf8-bytevector
    osi-port-count
+   osi-port-create-time
+   osi-port-name
+   osi-port?
```
owaddell: Would it make sense to provide an `osi-port-closed?` predicate? I'm not convinced it's worth doing. The obvious implementation would tell us only about the status of `handle`, not the underlying `fd`.

src/swish/osi.c
```diff
@@ -239,6 +233,19 @@ static ptr close_fs_port(uptr port, ptr callback) {
   return Strue;
 }
 
+static ptr close_fd_port(uptr port, ptr callback) {
+  fs_port_t* p = (fs_port_t*)port;
+  free(p);
+  osi_add_callback1(callback, Sfixnum(0));
```
owaddell: Clever how our `.close` vtable entries avoid unnecessary lock/unlock of the `callback`. I hadn't noticed that before doing a double-take here.

doc/swish/db.tex
```diff
@@ -452,6 +452,33 @@
...
+\var{x} is a database record instance.
+
+\defineentry{database-create-time}
+\begin{procedure}
+  \code{(database-create-time \var{db})}
+\end{procedure}
+\returns{} a clock time in milliseconds
+
+The \code{database-create-time} procedure returns the clock time from
+\code{erlang:now} when database record instance \var{db} was created.
```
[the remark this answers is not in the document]
burgerrg: We use this language several times in erlang.tex, so I think it should be adequately clear.

doc/swish/erlang.tex
```diff
@@ -1394,6 +1405,27 @@ \subsection{Miscellaneous}
...
+% ----------------------------------------------------------------------------
+\defineentry{make-osi-input-port}
+\begin{procedure}
+  \code{(make-osi-input-port \var{p})}
+\end{procedure}
+\returns{} a binary input port
+
+The \code{make-osi-input-port} procedure returns a custom binary input
+port that reads from osi-port \var{p}. It does not track position.
+
```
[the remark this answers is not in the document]
burgerrg: I like the idea of specifying the close behavior. Because not all osi-ports have underlying file descriptors, I'd prefer to mention that the underlying file descriptor is not closed in the documentation for open-fd-port.

src/swish/io.ss
```diff
@@ -349,9 +367,7 @@
   ;; Console Ports
 
   (define (make-console-input)
-    ;; no need to register static stdin port with the guardian
-    (define name "stdin-nb")
-    (binary->utf8 (make-iport name (make-osi-port name (erlang:now) (osi_get_stdin)) #f)))
+    (binary->utf8 (make-osi-input-port (open-fd-port "stdin-nb" 0 #f))))
 
   (define hook-console-input
```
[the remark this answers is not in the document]
burgerrg: Because Swish uses stderr as the mode of last resort when reporting problems, I would prefer not to hook it. I think that hooking stdout might be OK, but it's already dangerous to write to stdout from multiple processes at the same time.

src/swish/io.ss
burgerrg: Yes, good idea.

doc/swish/erlang.tex
```diff
@@ -1394,6 +1405,27 @@ \subsection{Miscellaneous}
 for the file \var{path} using \var{mode}, which defaults to
 \code{\#o777}. It returns \var{path}.
 
+% ----------------------------------------------------------------------------
```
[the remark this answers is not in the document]
burgerrg: Good ideas! I updated the osi-port functions in io.ss to raise an osi-port-closed tuple instead of the unhelpful invalid foreign-procedure argument error.

src/swish/osi.c
```diff
@@ -1095,7 +1100,7 @@ static void get_callbacks_timer_cb(uv_timer_t* handle) {
 
 ptr osi_get_callbacks(uint64_t timeout) {
   uv_update_time(osi_loop);
-  if (0 == timeout) {
+  if ((0 == timeout) || (Snil != g_callbacks)) {
```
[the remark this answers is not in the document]
burgerrg: I did not observe any bad behavior but thought to add this when I was writing the new code and didn't want to create a callback just to get the event loop to fire.

src/swish/osi.c
burgerrg: Yes, because the callback list is held in Scheme's heap, we don't need the lock!

src/swish/osi.c
```diff
@@ -998,6 +1005,17 @@ ptr osi_list_uv_handles(void) {
   return ls;
 }
 
+ptr osi_open_fd(int fd, int close) {
+  if (fd < 0)
+    return osi_make_error_pair("osi_open_fd", UV_EINVAL);
+  fs_port_t* port = malloc_container(fs_port_t);
+  if (!port)
+    return osi_make_error_pair("osi_open_fd", UV_ENOMEM);
+  port->vtable = close ? &file_vtable : &fd_vtable;
```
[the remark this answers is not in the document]
burgerrg: I wondered about this, too. Unfortunately on Windows there is no equivalent to fstat, so I'd prefer to leave things as they are, lest we break the redirected stdio handles in Windows. I did intend to allow file descriptors from outside Swish to work.

src/swish/io.ss
```diff
@@ -237,6 +266,11 @@
...
   (define osi-port-table (make-weak-eq-hashtable))
 
+  (define (osi-port-closed? p)
+    (unless (osi-port? p)
+      (bad-arg 'osi-port-closed? p))
+    (not (osi-port-handle p)))
+
   (define (osi-port-count) (hashtable-size osi-port-table))
 
   (define print-osi-ports
```
owaddell: It's not safe to do `(< (osi-port-handle p1) (osi-port-handle p2))` on line 287, is it? Less concerning, but related is line 292.

We disable interrupts only while grabbing the keys in the `osi-port-table` on line 280. While we're sorting the vector someone else could come in and call `close-osi-port` on one of the ports and `<` would be sad about `#f`. Looks like a race that would be difficult to reproduce.

burgerrg: Thank you for pointing this out! I fixed it and the other print routines that looked at handles when interrupts are active.

src/swish/db.ss
```diff
@@ -484,11 +488,9 @@
      (match (osi_prepare_statement* (database-handle db) sql
```
owaddell: Looks like line 480 should be storing `handle` now instead of 0 where we restore `db` on failure to close.

src/swish/db.ss
```diff
@@ -341,31 +348,36 @@
 
   (define (database-count) (hashtable-size database-table))
 
+  (define (sorted-cells ht create-time)
+    (let ([v (with-interrupts-disabled (hashtable-cells ht))])
```
owaddell: Nice! I didn't know about `hashtable-cells`.

src/swish/db.ss
burgerrg: Nice find, thanks!

PR #26, opened 2019-04-11
fix submodule check and some errors reported by valgrind

owaddell:
This pull request fixes the make rule that checks for out-of-date git submodules. We get different results depending on whether we run it from the top level of the repository or from the src/swish directory. The latter seems to always report that we're up-to-date.

This also tries to address some errors reported by valgrind:
1. libuv isn't populating the `st_flags` and `st_gen` fields, except on the mac
2. it looks like we're supposed to call `uv_close` on the `uv_process_t` instead of freeing it directly.

To see the valgrind issues, try:
```
$ git checkout master
$ ./configure && make
$ valgrind ./build/release/bin/swish
> (spawn-os-process "this-does-not-exist" '() self)
```

### DIFF
--- a/src/swish/osi.c
+++ b/src/swish/osi.c
@@ -804,8 +804,7 @@ ptr osi_spawn(const char* path, ptr args, ptr callback) {
     uv_close(&(in_port->h.handle), close_stream_cb);
     uv_close(&(out_port->h.handle), close_stream_cb);
     uv_close(&(err_port->h.handle), close_stream_cb);
-    Sunlock_object(callback);
-    free(p);
+    uv_close((uv_handle_t*)p, close_handle_data_cb);
     return osi_make_error_pair("uv_spawn", rc);
   }
   ptr r = Smake_vector(4, 0);
--- a/src/swish/Mf-base
+++ b/src/swish/Mf-base
@@ -1,12 +1,14 @@
-.PHONY: all build-dirs clean platform-clean ready reminders
+.PHONY: all build-dirs clean platform-clean ready reminders submodules
 
-SETUP := Makefile build-dirs git.revision chezscheme.revision
+SETUP := Makefile submodules build-dirs git.revision chezscheme.revision
 CORE := ../../${BUILD}/bin/swish${EXESUFFIX} ../../${BUILD}/bin/swish.boot ../../${BUILD}/bin/swish.library
 AUXLIB := ../../${BUILD}/lib/swish/mat.so ../../${BUILD}/lib/swish/profile.so ../../${BUILD}/lib/swish/testing.so
 
 ready:: | ${SETUP}
 ready:: io-constants.ss ${SwishLibs}
-	@if git submodule status --recursive | grep -q '^[^ ]'; then \
+
+submodules:
+	@if cd ../.. && git submodule status --recursive | grep -q '^[^ ]'; then \
 	  echo "Please update git submodules (or stage submodule change)."; \
 	  exit 1; \
 	fi


src/swish/osi.c
```diff
@@ -312,8 +312,13 @@ static void stat_cb(uv_fs_t* req) {
     Svector_set(r, 8, Sunsigned64(req->statbuf.st_size));
     Svector_set(r, 9, Sunsigned64(req->statbuf.st_blksize));
     Svector_set(r, 10, Sunsigned64(req->statbuf.st_blocks));
+#if defined(__APPLE__)
     Svector_set(r, 11, Sunsigned64(req->statbuf.st_flags));
     Svector_set(r, 12, Sunsigned64(req->statbuf.st_gen));
+#else
+    Svector_set(r, 11, Sfixnum(0));
+    Svector_set(r, 12, Sfixnum(0));
+#endif
```
burgerrg: libuv is responsible for setting its structures, and it appears to do so. See the uv__to_stat function in lib/src/unix/fs.c. I ran gcc -E on this file and found the assignment of 0 to st_flags and st_gen.

valgrind didn't complain when I took out this change on Ubuntu.

owaddell: I think there actually is a problem in libuv/src/linux/fs.c. Note that `uv__fs_stat` first calls `uv__fs_statx` and returns the result from that *without* going through `uv__to_stat`. It falls back on `stat` and `uv__to_stat` only when `uv__fs_statx` returns `UV_ENOSYS`. So it looks like valgrind is catching a bug.

I'm not thrilled with patching around it in our code, of course.

I see this on an x86_64 Fedora 29 box, where I'm using Valgrind 3.14.0 (vs. 3.13.0 on Ubuntu) and there is no error message about syscall 332 as there is on Ubuntu:
```
$ uname -r
5.0.6-200.fc29.x86_64
$ valgrind ./build/release/bin/
libosi.so  swish      
🔹[owaddell@toolbox swish]$ valgrind ./build/release/bin/swish 
==13313== Memcheck, a memory error detector
==13313== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==13313== Using Valgrind-3.14.0 and LibVEX; rerun with -h for copyright info
==13313== Command: ./build/release/bin/swish
==13313== 
==13313== Conditional jump or move depends on uninitialised value(s)
==13313==    at 0x4BF57D: Sunsigned64 (in /home/owaddell/tmp/swish/build/release/bin/swish)
==13313==    by 0x41401C: stat_cb (osi.c:315)
==13313==    by 0x5282B0: uv__work_done (threadpool.c:313)
==13313==    by 0x52C386: uv__async_io (async.c:118)
==13313==    by 0x53994E: uv__io_poll (linux-core.c:379)
==13313==    by 0x52CB7D: uv_run (core.c:363)
==13313==    by 0x41709C: osi_get_callbacks (osi.c:1103)
==13313==    by 0x9328832: ???
==13313==    by 0x4CEE04F: ???
==13313==    by 0x4CCCA0: S_call_help (in /home/owaddell/tmp/swish/build/release/bin/swish)
==13313==    by 0x4CE4DF: boot_call (in /home/owaddell/tmp/swish/build/release/bin/swish)
==13313==    by 0x4CE6E5: load (in /home/owaddell/tmp/swish/build/release/bin/swish)
[...]
> (software-revision)
((chezscheme . "19ebfa2b91f2b48ecc7cf2f186fc4b8c1203d0a9")
  (swish . "b08a3e07b1b47440599f9e3d1b091fa84cadb011"))
```

On the x86_64 Ubuntu 18.0.1 box, with Valgrind 3.13.0, I get:
$ uname -r
4.15.0-45-generic
$ valgrind ./build/release/bin/swish 
==10041== Memcheck, a memory error detector
==10041== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==10041== Using Valgrind-3.13.0 and LibVEX; rerun with -h for copyright info
==10041== Command: ./build/release/bin/swish
==10041== 
--10041-- WARNING: unhandled amd64-linux syscall: 332
--10041-- You may be able to write your own handler.
--10041-- Read the file README_MISSING_SYSCALL_OR_IOCTL.
--10041-- Nevertheless we consider this a bug.  Please report
--10041-- it at http://valgrind.org/support/bug_reports.html.
```
According to [this message](https://www.mail-archive.com/kde-bugs-dist@kde.org/msg221482.html) the unhandled 332 is `statx`. In libuv/src/unix/fs.c it looks like `uv__fs_stat` first tries `uv__fs_statx`, which calls `statx`, and otherwise falls back on `stat`.

Comment (burgerrg) [reply src/swish/osi.c]: Please enter an issue or, if you know how to fix it, a pull request to libuv.

In the meantime, why not simply add the following line just before the two places that call the stat functions in osi.c:
<pre>req->statbuf = (uv_stat_t){0};</pre>

Comment (owaddell) [reply src/swish/osi.c]: I entered [a libuv issue](https://github.com/libuv/libuv/issues/2262).

I'm not sure what to think of the proposed simplification.
1. It would mask libuv regressions.
1. It's not as ugly as the `#ifdef`, so we might forget to remove it when/if libuv fixes the issue.
1. It adds some overhead to our `stat` calls. For example, on a6le we add:
```
	movq	-8(%rbp), %rax
	leaq	112(%rax), %rdx
	movl	$0, %eax
	movl	$20, %ecx
	movq	%rdx, %rdi
	rep stosq
```

burgerrg: Thanks for submitting the libuv pull request! All the reviewers approved it, so now it should at some point be merged into the main branch. I got a pull request for documentation approved last year and needed to ping the group to get it merged.

burgerrg: I'd like to get the other changes into the master branch. What do you think of creating an issue to track the libuv pull request and then finish this pull request without the commit that works around it?

owaddell: Sounds good. Opened #27 to track the pull request.

owaddell:
Thanks, Bob!

burgerrg:
Thanks for finding and fixing these problems!

PR #32, opened 2019-04-19
Standardize application startup and shutdown

owaddell:
This pull request aims to make application startup and shutdown more robust and consistent across scripts and applications.

In particular, we now set `scheme-start` so we call the application entry point after the base system is fully initialized, as Kent outlined in a comment on https://github.com/cisco/ChezScheme/issues/383#issuecomment-455451479.

(As part of testing bfe0169579a5994072fce4be494b13f4bd32d8a1 I've optimistically rebased onto that commit.)


owaddell:
@burgerrg I just found that `app:path` is broken for a stand-alone app run via `$PATH`, so more changes are coming.

burgerrg:
Should you document the new swish-start procedure?

burgerrg:
In compile.ss, you might put (reset-handler abort) as the first expression. That way, if anything below goes wrong, Scheme will exit with a non-zero exit code.

burgerrg:
I don't think the new-cafe at the end of compile.ss is necessary. compile.ss is loaded from scheme and swish -q --, which go into a cafe after the file is loaded.

burgerrg:
I tweaked compile.ss so that it exits with the result of new-cafe. It avoid the double repl when things go well.

src/swish/app.ms
```diff
@@ -20,13 +20,157 @@
 ;;; OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 ;;; DEALINGS IN THE SOFTWARE.
 
+(include "swish/script-testing.ss")
```
burgerrg: Why this include since it will be imported a few lines down?

burgerrg: Perhaps Scheme won't find the library unless it's included?

burgerrg: Probably worth a comment explaining that.

owaddell: I removed the `include` by setting `library-directories` so that we find the testing library.

owaddell:
I agree that writing documentation is an important part of debugging the design.

One thing getting in my way is deciding where to document `swish-start`. Since we also need to document things like `app:name`, `app:path`, `app:start`, `app-sup-spec`, and `app-exception-handler`, I deferred documenting `swish-start` in the original pull request. Maybe some of this goes under the operating system interface in Chapter 2 since it's about starting an application. Longer term it would be nice to add a User Guide.

The other thing holding me up is deciding what to think about some changes I held back originally. I pushed them as "WIP tip hand ..." so you can see what I was mulling over.

burgerrg:
It's fine with me to defer the documentation.

I can't think of a case where I'd write a swish application that would require multiple swish-starts, but your changes would make it compositional. Don't forget to guard set-random-seed so that it runs only once like hook-console-input.

burgerrg:
Maybe it doesn't matter if set-random-seed is run multiple times.

burgerrg:
I'm pleased with the changes you've submitted and would support integrating them into the master branch.

owaddell:
Just so I'm on the same page: are you saying you're okay including the changes for compositionality (suitably rebased, with tests beefed up)?

burgerrg:
Yes, I'm okay with the compositionally changes included or not, your choice.

owaddell:
What do you think about setting `random-seed` in `$swish-start` instead? (pushed a sketch)

burgerrg:
Do you think we should hook the console input in the same place?

owaddell:
That might make sense. Is there any reason we wouldn't want to hook console input in cases where we're running a script or stand-alone application?

burgerrg:
If an application really wants to block the whole system for a read, it can use (standard-input-port). Otherwise, it should use the async stdin.

owaddell:
Okay, I'll hook console input in `$swish-start`.

owaddell:
Not sure why github still thinks some checks haven't completed. Travis-CI seemed okay.

burgerrg:
> Not sure why github still thinks some checks haven't completed. Travis-CI seemed okay.
The link shows the tests succeeded. I'll give them a kick to see if that helps.

owaddell:
Thanks, Bob!

PR #34, opened 2019-04-23
add Travis-CI testing on a6nt and i3nt

burgerrg:


src/vs
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
owaddell: Can you `chmod 755` this file so it works under cygwin as well?

src/swish/web-mats/static/ok.html
```diff
@@ -1 +1 @@
-<html><body><p>ok</p></body></html>
+<html><body><p>ok</p></body></html>
```
owaddell: Why did we need to change the line endings of `ok.html` and `ok.txt`?

src/run-mats
```diff
@@ -20,14 +20,20 @@ case $# in
   *) fail;;
 esac
 
-if [ ! -d "${LIBDIR}" ]; then
-  echo "${this} expected to find $LIBDIR"
-  echo ""
+GITTOP="$(git rev-parse --show-toplevel)"
+case "$(uname -s)" in
+  MINGW64_NT-* | MSYS_NT-*)
+    GITTOP="$(cygpath -u "$GITTOP")"
```
owaddell: Interesting. I take it `git` is returning a Windows path on those systems?

src/vs
burgerrg: Yes, sorry about that! I forgot that the MSYS implementation doesn't preserve file attributes.

src/swish/web-mats/static/ok.html
burgerrg: The Travis-CI Windows machine's git implementation flips the line endings of these files to CRLF. I thought this would be the easiest way to deal with it.

src/run-mats
burgerrg: It is, which was quite surprising.

src/swish/web-mats/static/ok.html
burgerrg: Travis-CI sets git config --global core.autocrlf to true, which is a bug that's been reported to Travis-CI.

burgerrg: As a result, we could tweak the build script to turn this option off and reset the affected files. That's what I'm doing for Chez Scheme.

burgerrg:
Let's see if I can get Travis-CI to leave the line endings alone.

burgerrg:
Good idea about deleting the index, but it didn't work because the build.sh file is running and is affected by it.

burgerrg:
Thanks, Oscar!

owaddell:
Thanks again, Bob!

PR #35, opened 2019-04-25
test threaded Chez Scheme

burgerrg:


burgerrg:
Correct, we don't support multi-threaded programs, but we do have users that have only the threaded version of Chez Scheme installed on their systems and would like to use Swish. See a9722308da8330a64e8e7040647ff251f0159108.

owaddell:
I see.

What do you think about adding that disclaimer to README.md, maybe under the Design section?

burgerrg:
I like those suggestions, thank you!

owaddell:
Thanks, Bob!

PR #37, opened 2019-05-17
SQLite 3.28.0

burgerrg:
SQLite 3.28.0 was released on April 16, 2019, and has been stable for the past month.


owaddell:
Thanks, Bob!

PR #40, opened 2019-06-05
JSON objects now use symbols for keys

burgerrg:
@owaddell, this will let you see what I had in mind. The json:make-object form now takes literal keys, which is more restrictive than before, but it seemed to affect only one piece of code. See what you think!


burgerrg:
The first commit fixes a parser bug that allowed for a trailing comma in arrays and objects.
The second commit commonizes the string output port used in json:read for a significant memory savings.
The third commit uses symbols for keys and symbol-hashtables for objects.

PR #42, opened 2019-06-28
Abstract foreign-handle management and reporting

owaddell:
This pull request aims to simplify the job of managing foreign handles.

I'm open to suggestions for improving the names. We could swap "handle" for "type", but "handle" may look more like a verb in context. I'm holding off on the documentation until we settle on the names and iron out the interface.

It might be cleaner to split out the non-WIP commits and move them through the process sooner.


src/swish/io.ss
```diff
@@ -265,53 +269,126 @@
...
+  (define (type-guardian-counts make-it set-it!)
+    (let ([it (make-it)])
+      (vector-for-each
+       (lambda (cell)
+         (set-it! it (car cell) ((%type-reporter-count (cdr cell)))))
+       (hashtable-cells type-reporters))
+      it))
+
+  (define (type-guardian-count type)
+    (match (hashtable-ref type-reporters type #f)
```
burgerrg: Suggestion: match-let* is more efficient when there is only one clause to match.

burgerrg: The same goes for type-guardian-print.

owaddell: Good idea. I modified the single-clause case of `match` to expand as `match-let*` since I often forget this and sometimes prefer the indentation of `match`. Hopefully I haven't missed anything in the transformation.

owaddell:
I renamed the new procedures and added documentation and tests. I plan to leave the commits split out this way to make the history easier to assimilate.

doc/swish/db.tex
```diff
@@ -117,12 +117,12 @@
...
   When a SQL string is not found in the cache,
-  \code{PrepareStatement} is used with the
+  \code{osi\_prepare\_statement} is used with the
   \code{current-database} to make a SQLite statement. The raw
   statement handle is stored in a Scheme record:\newline
   \code{(define-record-type statement (fields (immutable handle)
-    (immutable database)))}\newline The \code{statement} record is
-  not registered with a guardian. The statement is finalized using
+    (immutable database)))}\newline
+  The statement is finalized using
```
burgerrg: FinalizeStatement => osi\_finalize\_statement
CloseDatabase => osi\_close\_database
BindStatement => osi\_bind\_statement
StepStatement => osi\_step\_statement
ResetStatement => osi\_reset\_statement
GetStatementColumns => osi\_get\_statement\_columns

Thanks for finding these renames I missed when porting from Swish 1!

doc/swish/erlang.tex
```diff
@@ -758,6 +760,128 @@ \subsection{Miscellaneous}
...
+\end{procedure}
+\returns{} unspecified
+
+The \code{print-foreign-handles} procedure prints information about open
+foreign handles to textual output port \var{op}, which defaults to the current
+output port.
+The \code{print-foreign-handles} procedure
+calls the procedure returned by \code{(foreign-handle-print\ \var{type-name})}
+for each foreign-handle type registered with \code{make-foreign-handle-guardian}.
+
```
burgerrg: In arg-check below, the \ldots are indented too far. They should be at the name level as the let of arg.

web/swish/charts.ss
```diff
@@ -114,14 +116,23 @@
...
+(with-db [db (log-file) SQLITE_OPEN_READONLY]
+  (let* ([valid-charts
+          (append standard-charts
+            (map
+             (lambda (row)
+               (match row
+                 [#(,key)
+                  `(,key ,(format "json_extract(foreign_handles, '$.~a')" key))]))
+             (execute-sql db
+               "select distinct(key) from statistics, json_each(statistics.foreign_handles) order by key")))]
```
burgerrg: Does this query walk through the entire statistics table? If so, please try it on a full statistics table based on the automatic pruning.

burgerrg: If it's too slow, you could introduce a table to hold the foreign handle names and update the statistics table with an insert trigger that keeps this table up to date.

doc/swish/db.tex
owaddell: Good idea. Grep even turned up a hit in http.ms. That led to fixes for a few cases where we had `(receive (after n 'timeout) ...)` instead of `(receive (after n (raise 'timeout)) ...)`.

doc/swish/erlang.tex
owaddell: Good catch. Also fixed it so that `(arg-check 'foo)` doesn't result in a syntax error from `let`. Derived syntax might expand into an `arg-check` with no arguments.

web/swish/charts.ss
owaddell: Good catch!

I tried it with a full statistics table with one row every 5 minutes for 90 days. The original query was slow. When I changed it to grab the distinct keys via a subquery and moved the `order by` outside it sped up enough to be acceptable. The new query sped up even more when I changed it to operate on distinct `foreign_handles` so we skip redundant entries where we're at a steady state. I also tried it with an insert trigger along the way, but rolled it back once I found the `order by` mistake.

We'll eventually want to rewrite the diagnostics pages.

burgerrg: Great!

doc/swish/db.tex
owaddell: Travis pointed out that it no longer makes sense for http.ms to match the platform-specific error code from `uv_fs_open`.

owaddell:
Github seems to think I've missed some requested changes, but it's not clear what they are. I hit the re-request review button in case that was the step I missed. Looks like the Travis builds for i3le and ti3le ran into network connectivity issues before they could run the tests.

owaddell:
Reworded commit message for the documentation update the replaced old swish-win names with new osi names.

PR #45, opened 2019-07-12
A few small changes

owaddell:
I still need to document and test the `re` macro change, so left the last commit as WIP for now.


owaddell:
Added test and documentation.

PR #51, opened 2019-08-14
Add optional binary flag to open-file-to-*

owaddell:
@burgerrg are you open to a change like this? If so, I'll update the documentation and tests.

We've found that folks omit `O_TRUNC` by mistake when they meant to open a binary file to replace. I botched this in script-testing.ss, for example. If the `open-file-to-*` convenience procedures gave us the option of binary I/O, it might help folks avoid this mistake.


owaddell:
We're also considering the alternative of adding `open-binary-file-to-*` in place of the optional argument.

burgerrg:
I like adding the ability to open a binary file and think the binary? flag is a reasonable option.

burgerrg:
I also think it would be useful to pass an optional file mode to the open-file-* functions that need it, with a default of #o666 for non-Windows systems and #o777 for Windows.

burgerrg:
Or maybe just change the default #o777 to #o666 on non-Windows systems and have the user call the complete open-file for other cases.

src/swish/io.ss
```diff
@@ -689,32 +689,44 @@
...
+    (case-lambda
+     [(name) (open-file-to-read name #f)]
+     [(name binary?)
+      (open-file name O_RDONLY 0 (if binary? 'binary-input 'input))]))
+
+  (define open-file-to-write
+    (case-lambda
+     [(name) (open-file-to-write name #f)]
+     [(name binary?)
+      (open-file name (+ O_WRONLY O_CREAT O_EXCL) #o777
```
burgerrg: Use #o666 to be consistent with Chez Scheme. On Windows, it's ignored.

src/swish/io.ss
```diff
@@ -689,32 +689,44 @@
...
+        (if binary? 'binary-output 'output))]))
+
+  (define open-file-to-append
+    (case-lambda
+     [(name) (open-file-to-append name #f)]
+     [(name binary?)
+      (open-file name (+ O_WRONLY O_CREAT O_APPEND) #o777
```
burgerrg: Use #o666 to be consistent with Chez Scheme. On Windows, it's ignored.

src/swish/io.ss
```diff
@@ -689,32 +689,44 @@
...
+        (if binary? 'binary-append 'append))]))
+
+  (define open-file-to-replace
+    (case-lambda
+     [(name) (open-file-to-replace name #f)]
+     [(name binary?)
+      (open-file name (+ O_WRONLY O_CREAT O_TRUNC) #o777
```
burgerrg: Use #o666 to be consistent with Chez Scheme. On Windows, it's ignored.

src/swish/io.ss
owaddell: Good idea.

src/swish/io.ss
owaddell: Good idea.

src/swish/io.ss
owaddell: Good idea.

owaddell:
I opted for the `open-binary-file-to-*` variants that Chris suggested. It makes the intent clearer.

PR #52, opened 2019-08-16
Enable db gen-server support for URI filenames, in-memory, and temporary databases

laqrix:
- Updated to SQLite 3.29.0
- Enabled SQLITE_USE_URI compile-time option
- Changed db.ms to use in-memory databases throughout, except for the test where we test a variety of valid SQLite filenames

### DIFF
--- a/doc/swish/db.tex
+++ b/doc/swish/db.tex
@@ -50,8 +50,8 @@ transaction. In order to allow other processes access to the database,
 lazy transactions should be closed occasionally. To support this, it
 tracks a count of entries in the current transaction. A transaction is
 committed when the threshold of 10,000 is reached, the message queue
-of the \code{db} is empty, or when a direct transaction is
-requested.  Each database is created with write-ahead logging enabled
+of the \code{db} is empty, or when a direct transaction is requested.
+By default, each database is created with write-ahead logging enabled
 to prevent write operations from blocking on queries made from another
 connection.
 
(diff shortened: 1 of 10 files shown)


doc/swish/db.tex
```diff
@@ -249,6 +247,12 @@
   SQLITE\_OPEN\_READWRITE \code{SQLITE\_OPEN\_CREATE})}.
 \end{itemize}
 
+\var{db-init} is a procedure expects one argument, a database record
+instance. The return value is ignored. When the \var{mode} is
+\code{create}, and \var{filename} is not a special SQLite filename,
+the default procedure sets \code{journal\_mode} to ``wal'', otherwise,
+no additional initialization occurs.
+
```
burgerrg: \var{db-init} is a procedure **that takes** one argument, a database record
instance. The return value is ignored. When the \var{mode} is
\code{create} _{no comma here}_ and \var{filename} is not a special SQLite filename,
the default procedure sets \code{journal\_mode} to ``wal'' _{use semicolon here}_ otherwise,
no additional initialization occurs.

doc/swish/db.tex
```diff
@@ -141,16 +141,14 @@
...
-\code{journal\_mode} to ``wal'' if \var{mode} is
-\code{create}. The handle returned from \code{osi\_open\_database} is
-wrapped in a \var{database} record that is registered with a
-foreign-handle guardian using the type name \code{databases}.
-The foreign-handle guardian hooks
-the garbage collector so that dead databases are
-closed even if the \code{db} gen-server fails to close them for any
-reason.
+\genserver{db}{init} The \code{init} procedure takes a filename, mode
+symbol, and initialization procedure and attempts to open that
```
burgerrg: and **an** initialization procedure

laqrix: I debated that one when I wrote it. Updated.

doc/swish/db.tex
laqrix: I like that version better. Thank you. Updated.

laqrix:
Thanks Bob!

PR #53, opened 2019-08-19
libuv 1.32.0

burgerrg:
Upgrading to the latest stable libuv release. I'm not aware of any new functionality or fixed bugs that affect Swish.


owaddell:
Should we wait for 1.31.1?

I was about to open a PR here adding support for `uv_signal_start` and noticed that the libuv folks  approved a PR (2423) a few days ago that appears to fix a couple signal-handling issues. Guessing it might land as 1.31.1 before too long and it sounds as if there's no immediate benefit to jumping on 1.31.0 other than staying more current.

burgerrg:
Yes, let's wait for 1.31.1.

owaddell:
Looks like they jumped to v1.32.0.

PR #54, opened 2019-08-19
Misc updates

burgerrg:


NOTICE
```diff
@@ -3,9 +3,15 @@ Coulter, Inc. It is released under the terms of the MIT license.
 
 This product also includes separately licensed:
 
+* Chez Scheme by Cisco Systems, Inc.
+
 * jQuery by the jQuery Foundation
 
+* libuv
```
owaddell: Most other entries are of the form "xyz by organization". Is there anything useful for libuv that we should add?

burgerrg: I couldn't find an organization behind it. Since it's so well known, I think it's OK as is.

PR #56, opened 2019-08-21
fix "make coverage" tests to work with Chez Scheme 9.5.2

owaddell:
Updated a few other mats to work around an issue in the way
Chez Scheme 9.5.2 expands interleaved init expressions.
This will be fixed by f5fd9d144b68af26244855c7c2f34be97298deea
which is is expected to appear in Chez Scheme 9.5.4.

@burgerrg perhaps we should run "make coverage" in one or more of our travis builds?

### DIFF
--- a/.travis/test.sh
+++ b/.travis/test.sh
@@ -11,3 +11,4 @@ make test || {
     echo 'travis_fold:end:failures'
     exit $rc
 }
+make coverage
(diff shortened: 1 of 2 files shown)


burgerrg:
> @burgerrg perhaps we should run "make coverage" in one or more of our travis builds?

Great idea, I'm trying it now!

PR #57, opened 2019-08-22
fix swish-build for stand-alone program with -o filename.extension

owaddell:
The new test shows cases that were failing.

### DIFF
--- a/src/swish/swish-build
+++ b/src/swish/swish-build
@@ -276,6 +276,18 @@
 (define (change-extension fn new-ext)
   (string-append (path-root fn) new-ext))
 
+(define (platform-boot-file output-fn)
+  (string-append
+    (case (machine-type)
+      [(a6nt i3nt ta6nt ti3nt)
+       ;; On WIN32, Swish strips the .exe extension when looking for the boot
+       ;; file, as Chez Scheme does, so we strip .exe when creating the file.
+       (if (string-ci=? "exe" (path-extension output-fn))
+           (path-root output-fn)
+           output-fn)]
+      [else output-fn])
+    ".boot"))
+
 ;; to handle things like ".chezscheme.sls" -> ".chezscheme.so"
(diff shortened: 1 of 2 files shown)


burgerrg:
Chez Scheme strips off a trailing .exe in WIN32 when searching for a boot file (see lines 1104 ff. in c/scheme.c). Should we do the same thing here?

owaddell:
For WIN32, we strip off the .exe in run.c around line 108. Is that what you had in mind?

burgerrg:
Yes, the code in run.c looks good. I'm curious why the code in this pull request changes the extension for Windows and appends it for non-Windows.

owaddell:
I think `swish-build`removes the extension on Windows so that it creates a boot file with the name that run.c will end up looking for.

Perhaps I should tighten that up a bit by changing the extension on Windows only if it's `.exe`?

burgerrg:
> Perhaps I should tighten that up a bit by changing the extension on Windows only if it's `.exe`?

That's what I was thinking. Please add a comment saying so to remind us in the future.

src/swish/swish-build
```diff
@@ -276,6 +276,18 @@
 (define (change-extension fn new-ext)
   (string-append (path-root fn) new-ext))
 
+(define (platform-boot-file output-fn)
+  (string-append
+    (case (machine-type)
+      [(a6nt i3nt ta6nt ti3nt)
+       ;; On WIN32, Swish strips the .exe extension when looking for the boot
+       ;; file, as Chez Scheme does, so we strip .exe when creating the file.
+       (if (equal? "exe" (string-downcase (path-extension output-fn)))
```
burgerrg: Suggestion: (string-ci=? "exe" (path-extension output-fn))

owaddell: Good idea! I forgot all about `string-ci=?`.

PR #58, opened 2019-08-22
Improve process parameter performance

burgerrg:
Use a weak-eq-hashtable mapping the process-parameter procedure to the value in each process instead of a separate weak-eq-hashtable for each process-parameter.

### DIFF
--- a/doc/swish/erlang.tex
+++ b/doc/swish/erlang.tex
@@ -88,8 +88,9 @@ process receives it.
 A \emph{process} is an instance of \code{pcb}, a Scheme record type
 that extends \code{q} with an immutable \code{id} field, the process's
 unique positive exact integer, an immutable \code{create-time} field,
-the process's create time from \code{erlang:now}, and the following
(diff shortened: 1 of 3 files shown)


doc/swish/erlang.tex
```diff
@@ -88,8 +88,9 @@
 A \emph{process} is an instance of \code{pcb}, a Scheme record type
 that extends \code{q} with an immutable \code{id} field, the process's
 unique positive exact integer, an immutable \code{create-time} field,
-the process's create time from \code{erlang:now}, and the following
-mutable fields:
+the process's create time from \code{erlang:now}, an immutable
+\code{parameters} field, the process's weak eq-hashtable of process
```
owaddell: Consider "... weak eq-hashtable **mapping** process parameters to values ..." which is more like what we had at line 1109 before.

burgerrg: Good suggestion, thanks!

PR #61, opened 2019-08-27
http: let the handler deal with content when the type is not one the system handles

burgerrg:
No longer scoop all unhandled-content into a parameter but leave it to the handler. Whatever the handler doesn't read, the system ignores.

I added position tracking to most custom ports to support this change.


owaddell:
Chris and I pushed a proposal that
- adds some test coverage
- returns an internal server error if the handler violates the protocol by reading too much data

src/swish/http.ss
```diff
@@ -462,7 +467,7 @@
               (raise 'content-limit-exceeded))
             (let ([content (make-bytevector len)])
               (get-chunk! ip content len)
-              (values (parse-encoded-kv content 0 len) #f))]
+              (values (parse-encoded-kv content 0 len) pos))]
```
burgerrg: Since we just read exactly len bytes, I don't see a need to return pos here.

src/swish/http.ss
```diff
@@ -475,14 +480,14 @@
...
-             (define-values (params pos)
-               (http:get-content-params&pos ip header))
-             (on-exit (delete-tmp-files params)
-               (http:file-handler ip op request header
-                 (append query params)))
-             (when (keep-alive? header)
-               (advance! ip pos)
-               (http:handle-input ip op))]))]
+             (define opos (port-position op))
+             (let-values ([(params ipos) (http:get-content-params&pos ip header)])
```
burgerrg: Let's keep it in define and define-values form to avoid the indentation.

burgerrg:
I'm happy to incorporate these changes and update the documentation if you'd like.

src/swish/http.ss
owaddell: Returning `pos` lets us detect another case where the handler reads too much. See line 277 of http.ms. If the handler behaves, `advance!` is pretty low overhead.

Aside: perhaps we should pull the `eof-object?` test outside the loop in `advance!`:
```
(do ([n remaining (fx- n 1)] [b #f (get-u8 ip)])
    ((fx= n 0)
     (when (eof-object? b)
       (raise 'unexpected-eof))))
```

src/swish/http.ss
owaddell: Have you seen what `define-values` expands to? It negates any performance benefit we may be trying to obtain here via multiple return values.

If we always return `pos` when content length is provided, we may be able to simplify the code and eliminate the use of multiple return values.

owaddell: I pushed a sketch that may help w/ indentation, but I haven't compared the generated code.

owaddell:
Thanks, Bob! I'd be happy for you to incorporate the changes and update the documentation (and squash the commits Chris and I tacked on).

src/swish/http.ss
burgerrg: OK, I'll put the pos back in there. I chose not to pull the eof-object? test because it's very quick and this code should not run at all in the common case.

src/swish/http.ss
burgerrg: Thanks, I didn't realize what define-values generated. I'll keep the let-values.

src/swish/http.ss
```diff
@@ -425,30 +425,50 @@
...
+               [type (or (http:find-header "Content-Type" header) "none")]
+               [pos (+ (port-position ip) len)])
+          (cond
+           [(multipart/form-data-boundary type) =>
+            (lambda (boundary)
+              (return
+               (parse-multipart/form-data ip (string-append "--" boundary))
+               pos))]
+           [(starts-with? type "application/x-www-form-urlencoded")
+            (when (> len (http-content-limit))
```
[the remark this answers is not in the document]
burgerrg: I updated the documentation to be more precise about the use of this limit. I don't see a need for additional tests. Code inspection is fine.

src/swish/http.ss
```diff
@@ -425,30 +425,50 @@
...
-     [else '()]))
+        (let* ([len (or (string->number x)
+                        (raise `#(invalid-content-length ,x)))]
+               [type (or (http:find-header "Content-Type" header) "none")]
+               [pos (+ (port-position ip) len)])
+          (cond
+           [(multipart/form-data-boundary type) =>
+            (lambda (boundary)
+              (return
+               (parse-multipart/form-data ip (string-append "--" boundary))
```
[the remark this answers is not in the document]
burgerrg: It seems to work quickly enough as is, so I'll leave it.

burgerrg:
I've made the requested changes and updated the documentation. In the process I noticed a few error expressions from http.ss that were not in errors.ss, so I added them.

doc/swish/http.tex
```diff
@@ -195,8 +197,13 @@
 \hasvalue{} a positive fixnum
 
 The \code{http-content-limit} parameter specifies the maximum length
-in bytes of the HTTP request content excluding uploaded files. It
-defaults to 4,194,304.
+in bytes of HTTP request content read by the system handlers when
+\code{Content-Type} is \code{multiform/form-data} or
```
owaddell: multiform => multipart

burgerrg: Thanks for catching this!

PR #62, opened 2019-08-29
add signal-handler

owaddell:
Note that the forthcoming libuv 1.31.1 includes a signal-handling fix.


laqrix:
Open question, should we hook SIGTERM (maybe SIGHUP) by default to call (app:shutdown) so that apps built upon Swish have a good default shutdown behavior?

burgerrg:
Yes, I think by default we should hook SIGTERM on posix systems and SIGINT, SIGBREAK, and SIGHUP on Windows systems.

The posix systems can handle SIGINT by calling reset, but for some reason CTRL-C usually causes the Windows repl to exit. I'm not sure why it behaves differently, but it may have to do with the console read getting clobbered by CTRL-C. In posix, the read returns EINTR and is retried.

burgerrg:
A natural place to put the default handler for SIGTERM is $swish-start, right before calling run.

burgerrg:
The default SIGTERM handler should be `(lambda () (spawn app:shutdown))`.

src/swish/io.ms
```diff
@@ -586,4 +587,105 @@
    'ok)
   )
 
+(isolate-mat signal-handler ()
+  (match-let*
+   ([#(EXIT #(bad-arg signal-handler sym)) (catch (signal-handler 'sym))]
+    [#(EXIT #(bad-arg signal-handler "string")) (catch (signal-handler "string" #f))]
+    [#(EXIT #(bad-arg signal-handler 123)) (catch (signal-handler SIGHUP 123))]
```
burgerrg: Use SIGWINCH instead of SIGHUP since SIGHUP is trapped by default on Windows.

doc/swish/osi.tex
```diff
@@ -422,6 +430,26 @@
 \var{pid}. It returns \code{\#t} when successful and an error pair
 otherwise.
 
+\defineentry{osi\_signal\_start}
```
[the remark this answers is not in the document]
owaddell: Oops! Thanks for catching this.

Aside: it would be nice to fix our LaTeX setup so we don't have to escape underscores in these cases. It'll make grep easier.

src/swish/app-core.ss
```diff
@@ -125,6 +126,18 @@
...
+    (spawn application:shutdown))
+
+  (define (trap-signals)
+    (meta-cond
+     [(memq (machine-type) '(i3nt ti3nt a6nt ta6nt))
+      (signal-handler SIGBREAK handle-signal)
+      (signal-handler SIGHUP handle-signal)
+      (signal-handler SIGINT handle-signal)]
+     [else
+      (signal-handler SIGTERM handle-signal)]))
```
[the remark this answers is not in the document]
owaddell: Good call!

src/swish/io.ms
owaddell: Good plan.

src/swish/app.ms
```diff
@@ -199,6 +199,72 @@
...
+     (["started"
+       (receive (after 1500 'bad-start)
+         [(,@from-stdout . ,msg) msg])]
+      [,!
+       (meta-cond
+        [(memq (machine-type) '(i3nt ti3nt a6nt ta6nt))
+         ;; osi_kill won't work for this on Windows, but I haven't
+         ;; gotten GenerateConsoleCtrlEvent to work because I gave
+         ;; it a dwProcessGroupId that tore down too much.
+         (raise 'need-help-from-Bob)]
```
owaddell: @Bob, is there some way to automate this test or will we have to test this manually?

owaddell:
Thanks for the suggestions, Bob!

src/swish/app.ms
burgerrg: I don't know how to automate this test on Windows, so I'd leave it as a manual test.

burgerrg:
My change to io.ms didn't fix it on Windows all the time. The handle sort order isn't always as expected. I think we should rewrite the test to split out the lines from the signal-handler print, sort them by signal number, and then compare against the expected set.

owaddell:
I updated io.ms as recommended. Also took a stab at a manual test for Windows.

burgerrg:
We should add a paragraph, perhaps at the end of the signal-handler entry in erlang.tex, saying that the system puts handlers in place for various signals depending on the platform that call app:shutdown safely.

owaddell:
Good idea!

burgerrg:
Looks good, please rebase and merge to master at will!

PR #63, opened 2019-09-04
Restore interactive Ctrl-C behavior

owaddell:
Quick sketch. Can we do something better?

```
$ swish
> (let f () (f))
^C
break> i
#<system continuation>              : q

break> q
>
```

### DIFF
--- a/doc/swish/erlang.tex
+++ b/doc/swish/erlang.tex
@@ -1481,6 +1481,16 @@ operation. To unblock the process, call \code{complete-io} from a
 callback function.
 
 % ----------------------------------------------------------------------------
+\defineentry{windows?}
+\begin{syntax}
+  \code{windows?}
+\end{syntax}
+\expandsto{} a boolean
+
+The \code{windows?} macro expands to \code{\#t} if the host is running
+Microsoft Windows and \code{\#f} if not.
+
+% ----------------------------------------------------------------------------
 \subsection {Tuples}
 
 For users of the concurrency model, a \emph{tuple}\index{tuple} is a
--- a/doc/swish/erlang.tex
+++ b/doc/swish/erlang.tex
@@ -105,8 +105,9 @@ process parameters to values, and the following mutable fields:
   dead
 \item \code{precedence}: wake time if sleeping or 0 if ready to run
 \item \code{flags}: fixnum with bit 0 set when sleeping, bit 1 set
-  when the process traps exits, and bit 2 set when the process is
-  blocked for I/O
+  when the process traps exits, bit 2 set when the process is
+  blocked for I/O, and bit 3 set when the process is pending a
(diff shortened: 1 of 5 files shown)


burgerrg:
CTRL-C doesn't work reliably in a Swish system because the signal may arrive when any process is running. The keyboard-interrupt-handler calls reset, which doesn't work well in processes other than the repl. I suppose you could use the new signal-handler to locate the repl process and hack its continuation so that it calls reset when it comes back on the run queue. Sounds dangerous.

burgerrg:
new-cafe parameterizes $interrupt so that it prints the break message and enters a new loop that parameterizes reset. Somehow you'll have to get the signal to call the keyboard-interrupt-handler from that Swish process, otherwise it won't do what you expect.

burgerrg:
I extended your approach by adding an interrupt? flag to the pcb and use it to call keyboard-interrupt from the interactive repl. It doesn't interrupt console input, so if you press CTRL-C during input, you'll need to press enter before the effects take place. It almost works on Windows, too, but CTRL-C seems to mess up standard input there so that the repl exits.

owaddell:
Nice!

I pushed a sketch that spawns one process to read from the real console and another process that sends the output to the "current" console-input-port. It's a rough sketch, but seems to work for the following cases on Linux. I haven't tried building it on Windows.

#### ^C at the repl
```
$ swish
> ^C
> ^C
> (+ 2 3)
5
>
```

#### ^C in receive
```
$ swish
> (receive)
^C
break> e
^C
break> r
>
```

#### ^C in receive, with new-cafe in break
```
$ swish
> (receive [foo 'bar])
^C
break> n
>> (send self 'foo)
>> ^D
break> e
bar
```

#### ^C to break out of loop
```
$ swish
> (let f () (f))
^C
break> i
#<system continuation in dynamic-wind>           : q
break> e
^C
break> r
> 123
123
```

#### ^C during read
```
$ swish
> (define x (read))
^C
break> e
^C
break> e
howdy
> x
howdy
```

#### new-cafe at repl
```
$ swish
> (new-cafe)
>> (+ 2 3)
5
>> ^D
> cdr
#<procedure cdr>
>
```

burgerrg:
The separate process idea worked out very well for Unix-like systems where the signal is delivered in the dynamic extent of the read call (and its EINTR retries in libuv and Swish if libuv misses).
On Windows, CTRL-C causes the read to return 0 as though CTRL-Z were pressed, and it takes the separate thread a while to deliver the asynchronous signal.
Interactive CTRL-C now works for all platforms. Unfortunately, Cygwin calls swish from a forked bash process that handles CTRL-C by exiting, so you'll have to use CMD on Windows.

burgerrg:
Sometimes CTRL-C works for me in Cygwin.

PR #64, opened 2019-09-05
swish-build now sets the executable bit on linked scripts

burgerrg:


burgerrg:
Thanks, I updated it to use O_TRUNC and binary-output. I also added a commit to update NOTICE.

burgerrg:
Oops, I undid the sorting in NOTICE...

burgerrg:
OK, I updated NOTICE to say libuv by the libuv team, since that is what they call themselves on libuv.org.

PR #65, opened 2019-09-05
HTTP paths must start with /

burgerrg:
The code should check for absolute paths and then remove the leading / when forming the full path.

### DIFF
--- a/src/swish/http.ss
+++ b/src/swish/http.ss
@@ -161,17 +161,17 @@
           state
           (let ([state (load-watchers state)])
             ($state copy [mime-types (read-mime-types)]))))
-    (define (url->abs-paths path)
-      (map
-       (lambda (x) (path-combine (web-dir) x))
-       (cond
-        [(char=? (string-ref path (- (string-length path) 1)) #\/)
-         (list (string-append path "index.ss"))]
-        [(string=? (path-extension path) "")
-         (list path
-           (string-append path ".ss")
-           (path-combine path "index.ss"))]
-        [else (list path)])))
+    (define (url->abs-paths path) ;; path starts with "/"
+      (let ([path (path-combine (web-dir)
+                    (substring path 1 (string-length path)))])
+        (cond
+         [(directory-separator? (string-ref path (- (string-length path) 1)))
+          (list (string-append path "index.ss"))]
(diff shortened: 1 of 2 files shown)


src/swish/http.ss
```diff
@@ -161,17 +161,17 @@
...
-       (cond
-        [(char=? (string-ref path (- (string-length path) 1)) #\/)
-         (list (string-append path "index.ss"))]
-        [(string=? (path-extension path) "")
-         (list path
-           (string-append path ".ss")
-           (path-combine path "index.ss"))]
-        [else (list path)])))
+    (define (url->abs-paths path) ;; path starts with "/"
+      (let ([path (path-combine (web-dir)
```
laqrix: Rebinding path changes the behavior on Windows. (web-dir) is a native OS path, ending in \ on Windows.  If the path argument is "/", then this new path binding now ends in "\".  The original code stopped on the first cond clause, while the new code stops on the second clause -- thus yielding 3 proposed abs-paths, rather than 1.

burgerrg: Good catch! Maybe the first cond case should be (directory-separator? (string-ref ...)). That should fix it on Windows.

laqrix: It seems that the original path variable is in URL format, making "/" the only valid character. The newly introduced path variable could be a path in Windows, Unix, or mixed format.  I'm not sure directory-separator? is sufficient. We may need to test both "/" and "\".

burgerrg: Yes, the new path variable is formed by path-combine from the platform-specific path and the URL path in posix format. As a result, it cannot be the empty string (because of the way path-combine works). On Windows systems, directory-separator? checks for / and \\, which works well here. On posix systems, directory-separator? checks for / only. In all cases, it does what we want here.

laqrix: I didn't realize directory-separator? considered both / and \ on Windows.  Let's try that and make sure we get the right number of results from attempting the root '/" url.

burgerrg: I made the change, and it works on the root '/' url.

burgerrg: I also used string-append instead of path-combine since the directory-separator? test immediately preceding is exactly what path-combine does.

PR #66, opened 2019-09-09
handle errors with cyclic structures

burgerrg:
The coerce function in log-db.ss needs to enable print-graph to avoid warnings from cyclic structures, and the errors web page needs to enable print-graph when it processes these potentially cyclic structures.

### DIFF
--- a/src/swish/log-db.ms
+++ b/src/swish/log-db.ms
@@ -66,5 +66,9 @@
      (coerce
       (json:make-object
        [foo 123]
-       [bar (list 1 "two" (json:make-object [zip #t]))]))])
+       [bar (list 1 "two" (json:make-object [zip #t]))]))]
+    ["#0=(1 . #0#)"
+     (coerce '#0=(1 . #0#))]
+    ["#(error \"Exception occurred with condition components:\\n  0. &irritants: #0=(1 . #0#).\")"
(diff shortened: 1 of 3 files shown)


owaddell:
If we set `print-graph`, we'll get (less readable) graph output even in cases where there is no cycle.

We could set the combination of `print-length` and `print-level` instead, but maybe it's hard to pick reasonable values for those parameters.

burgerrg:
The graph annotations are used when there is shared structure. Since the goal of coerce is to faithfully represent a datum in the database, I think this is a good option. The errors page reads it back in and then converts to an English string.

One alternative is to catch the warning about cyclic structures and not let it crash the logger. I prefer making the shared structures explicit. It might be helpful in debugging, too.

owaddell:
Turning `print-graph` on is fine with me. As you say, it may help in debugging. It may even save space here and there. While `coerce` can't achieve the stated goal in general, it's good to preserve data where we can.

At some point we may want some bound on the size of the coerced result. If so, setting `print-level` and `print-length` may be preferable to simply truncating the output. It would provide a weak bound, but we would have some hope of reading back the representation with ellipses. We might be able to set these parameters via the new db-init parameter when we start log-db.

burgerrg:
Thanks, limiting the length is a good idea, too, because sometimes an error contains a very large data structure.

PR #68, opened 2019-09-17
Add support for running as a service on Windows and Linux

owaddell:
Note that the WIP commits will be deleted before the final pull request, I'm just including the service example so you can try it out more easily.

We added `osi_send_request` for the benefit of the Windows service code; it may be useful in other cases.

The travis updates are guesswork. We'll see how that plays out.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ matrix:
             - gcc-multilib
             - g++-multilib
             - lib32ncurses5-dev
+            - libsystemd-dev:i386
             - libx32ncurses5-dev
             - uuid-dev:i386
       before_script:
@@ -28,16 +29,25 @@ matrix:
             - gcc-multilib
             - g++-multilib
             - lib32ncurses5-dev
+            - libsystemd-dev:i386
             - libx32ncurses5-dev
             - uuid-dev:i386
       before_script:
         - sudo sysctl net.ipv6.conf.all.disable_ipv6=0
     - env: TARGET_MACHINE=a6le
       os: linux
+      addons:
+        apt:
+          packages:
+            - libsystemd-dev
       before_script:
         - sudo sysctl net.ipv6.conf.all.disable_ipv6=0
     - env: TARGET_MACHINE=ta6le
       os: linux
+      addons:
+        apt:
+          packages:
+            - libsystemd-dev
       before_script:
         - sudo sysctl net.ipv6.conf.all.disable_ipv6=0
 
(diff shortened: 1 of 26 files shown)


burgerrg:
uv_key_get on Windows requires a value that was previously set, so this function isn't a good way to tell if we're being called on Scheme's thread because it could be called from a foreign thread whose key wasn't set.

burgerrg:
Wow, uv_key_get calls abort() if it can't get the key. That's harsh!

burgerrg:
I think we should use pthread_self and GetCurrentThreadId to determine the running thread. Then we won't crash in Windows when called from a non-libuv thread.

burgerrg:
I pushed a commit that eliminates the use of thread-local storage to determine the scheme thread.

burgerrg:
The test-request.c code should copy the string passed from Scheme into the C heap. The lock-objects in the osi.ms should not be necessary. I'll update the code & tests.

burgerrg:
I pushed changes to the tests for osi_send_request to check the main-thread case and to use native threads instead of the libuv wrappers.

burgerrg:
I don't know why the 32-bit Linux osi.ms tests are failing.

burgerrg:
My "improved" test code had an off-by-one error in copy_string, and the call_now case added to the callback list outside of the run_callbacks site, which isn't garbage-collector safe. They're fixed now. I'm really glad the i3le tests caught these bugs in my tests!

burgerrg:
I think it would be wise to have osi_send_request fail with UV_EPERM when called from the main thread. The reason is that osi_add_callback is not safe to use except in the context of the uv event loop. I can't think of a case where I'd use osi_send_request from the main thread.

burgerrg:
I pushed changes:
- update the Linux packages in README.md
- allocate an extra NULL entry at the end of argv arrays to follow the specification
- osi_send_request now returns UV_EPERM when called from the main thread

owaddell:
@burgerrg thanks for the really helpful review (and for making the changes)! BTW, I really appreciated your thorough review of the signal-handling pull request as well.

burgerrg:
You're welcome! I spent a day on osi_send_request. I thought I could eliminate some of the complexity by having it build and send an async request, but I found that libuv isn't thread safe when creating async requests. Your solution worked well. Because the sending thread blocks until the return of the main thread processing, one can use local stack storage for the arguments, which is very handy.

owaddell:
One more build after removing the WIP commits, some of which touched non-example files.

owaddell:
Thanks again for the really helpful feedback and fixes, Bob!

PR #70, opened 2019-09-26
Add travis-ci configurations for the master branch Chez Scheme

burgerrg:


burgerrg:
Subsumed by https://github.com/becls/swish/pull/95.

PR #71, opened 2019-09-27
use LaTeX fonts in dot figures and fit to width

owaddell:
Chris pointed out that we had both pushed swish.pdf to gh-pages and wondered what the difference was. It looks like we getting different fonts in the `dot` output.

This is a sketch at using the LaTeX fonts in the figures so we generate the same output. I'm not 100% sure I've got the [fontconfig](https://www.freedesktop.org/software/fontconfig/fontconfig-user.html) stuff figured out, but this appeared to work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,7 @@ is not thread safe and should be used from the main thread only.
 - ginstall and realpath (can be installed through homebrew using `brew install coreutils`)
 - XCode and the command-line tools must be set
   (Xcode->Preferences->Locations->command line tools)
+- Latin Modern fonts from LaTeX (can be installed with Font Book from a location like `/usr/local/texlive/2019/texmf-dist/fonts/opentype/public/lm`)
 
 ## Windows
 
--- a/doc/.gitignore
+++ b/doc/.gitignore
@@ -17,5 +17,6 @@
 *.thm
 *.toc
 /c-coding-standard.pdf
+/font.conf
 /scheme-coding-standard.pdf
 /swish.pdf
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -11,8 +11,15 @@ all: $(pdf)
 %.pdf: %.tex sagian.sty swish.sty reference.bib
 	./run-latex $*
 
-%.pdf: %.dot
-	dot -Tpdf -o $@ $<
+%.pdf: %.dot font.conf
+	FONTCONFIG_FILE=$${PWD}/font.conf dot -Tpdf -o $@ $<
+
+font.conf:
+	@echo '<?xml version="1.0"?>' > $@
+	@echo '<!DOCTYPE fontconfig SYSTEM "fonts.dtd">' >> $@
+	@echo '<fontconfig>' >> $@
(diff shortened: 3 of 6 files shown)


burgerrg:
dot 2.40.1 on macOS doesn't look at font.conf at all (as determined by opensnoop), so it produces output in the wrong font. Any ideas how to get it to work as desired on macOS?

burgerrg:
If I used the font family name in the dot files and installed the LaTeX lm fonts in Font Book on macOS, then dot finds them. But this technique may not work on Linux or Windows.

owaddell:
For Linux and Windows we may be able to use the fontconfig mechanism. Maybe we can replace my `tt` and `roman` in the generated `font.conf` with the font family names you're using on macOS.

I tested the original commit on cygwin, Ubuntu, and Fedora and the fontconfig override seemed to work, but I'm not sure what font names you ended up having to use on the mac, so not sure what to try here.

burgerrg:
I used "Latin Modern Mono" and "Latin Modern Roman" in the dot files.

owaddell:
Hmm. Strangely enough, that's what I had originally (before adding the font.conf file) and I had trouble where it chose the Mono font for _both_ the mono and the roman, so I used the silly `tt` and `roman` names and added font.config `<alias>`.

I'm trying again. Still haven't found the right incantation in the font.conf file to get it to work. Seems like it the `<alias>` stuff shouldn't be needed.

owaddell:
Weird. When I remove the `<alias>` entries from font.conf and just leave in the `<dir>`, I get something sensible from `fc-match`:
```
$ export FONTCONFIG_FILE=${PWD}/font.conf
$ fc-match "Latin Modern Roman"
lmroman10-regular.otf: "Latin Modern Roman" "10 Regular"
$ fc-match "Latin Modern Mono"
lmmono10-regular.otf: "Latin Modern Mono" "10 Regular"
```
but, continuing in that shell, `dot` resolves both to mono:
```
$ dot -v -Tpdf -o swish/http-tree.pdf swish/http-tree.dot |& grep fontname:
fontname: "Latin Modern Mono" resolved to: (PangoCairoFcFont) "Latin Modern Mono, 10 Regular" /usr/share/texlive/texmf-dist/fonts/opentype/public/lm/lmmono10-regular.otf
fontname: "Latin Modern Roman" resolved to: (PangoCairoFcFont) "Latin Modern Mono, 10 Regular" /usr/share/texlive/texmf-dist/fonts/opentype/public/lm/lmmono10-regular.otf
```

owaddell:
Wow. I stumbled across a graphviz issue again and read enough of it this time to find that adding a comma to the end of the font name disables some unwanted font translation or other.

What I just pushed appears to work on Ubuntu, Fedora, and Cygwin. Diffpdf seemed happy enough with the output. See the commit message for details. Curious to see what happens on the mac.

burgerrg:
Thanks for figuring out how to get LaTeX's font in the diagrams produced by dot on the three platforms!!

PR #72, opened 2019-10-01
swish-test: add --info option to summarize software info from test results

owaddell:
Added an --info option and used it in run-mats script so Travis builds summarize software-info meta-data for test results. This gives a summary of the meta-data we'd normally see in the HTML report, which is not available to us on the Travis build.

This pull request also includes a fix for spurious errors when trying to use the `run-mat` script using recent csv9.5.3.


owaddell:
Oh, right. We current get "<unknown>" for the Chez Scheme revision because we build the 9.5.2 release and the change that provides revision info narrowly missed that release window.

owaddell:
Thanks for the clean-up, Bob!

I worked in a couple other minor fixes on this branch. Perhaps I should have opened separate pull requests.

PR #73, opened 2019-10-01
use sealed records when possible for greater efficiency

burgerrg:
Sealed record types perform record type checks very efficiently, so let's use them whenever possible.


burgerrg:
I noticed that db.tex had incomplete record definitions in there, so I updated them. Not all of the fields are documented.

PR #74, opened 2019-10-25
Fix bugs in close_fd_port and osi_get_bytes_used

burgerrg:
`close_fd_port` called `osi_add_callback1` outside of `osi_get_callbacks`, and this sometimes led to a panic with `#(event-loop-process-terminated #(bad-arg complete-io #<garbage>))`.

On macOS, the `malloc_zone_pressure_relief` call in `osi_get_bytes_used` sometimes causes an invalid memory reference, so I removed it.

### DIFF
--- a/src/swish/osi.c
+++ b/src/swish/osi.c
@@ -610,7 +610,6 @@ ptr osi_get_argv() {
 
 size_t osi_get_bytes_used(void) {
 #if defined(__APPLE__)
-  malloc_zone_pressure_relief(NULL, 0);
   struct mstats ms = mstats();
   return ms.bytes_used;
 #elif defined(__GLIBC__)
--- a/src/swish/osi.c
+++ b/src/swish/osi.c
@@ -22,6 +22,8 @@
 
 #include "osi.h"
 
+#include <assert.h>
+
 #ifdef _WIN32
 #pragma comment(lib, "advapi32.lib")
 #pragma comment(lib, "iphlpapi.lib")
@@ -101,6 +103,7 @@ static send_request_t g_send_request;
 static osi_thread_t g_scheme_thread;
 
 void osi_add_callback_list(ptr callback, ptr args) {
+  assert(g_callbacks); // must only be called via osi_get_callbacks
   g_callbacks = Scons(Scons(callback, args), g_callbacks);
 }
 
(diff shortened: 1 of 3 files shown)


burgerrg:
Nice finds!

owaddell:
The `osi_add_callback_list` constraint is a bit of a sharp edge. Is there a good way to help folks avoid or at least discover these kinds of errors?

I pushed a quick sketch at barking if `osi_add_callback_list` is called out of context. It may be more efficient to bark in `osi_get_callbacks` if `g_callbacks` is not Snil, but I thought catching it where we add callbacks might give folks a more useful stack if they trip on this. Hopefully branch prediction will mitigate the cost of the added check. Is there a better option? We could add an inspection checklist that includes this, but that isn't ideal.

This turned up a problem in shilbtest, so I took a really quick stab at sorting that out. Likely buggy.

burgerrg:
I had done something similar on Friday, so I tweaked your implementation to include the two other places that call uv_run and can therefore invoke callbacks. See what you think.

I'm happy with the assert in the code because adding to the callback list in the wrong context causes some very strange errors that are very difficult to debug.

I renamed the static g_callbacks to osi_callbacks so that run.c could use it. I decided against publishing it in swish.h because I don't want custom code using it.

burgerrg:
It looks as though the Windows C compiler can't link symbols like osi_loop across DLLs. I'll need to introduce an osi_get_loop() function for Windows and try again.

burgerrg:
No, Windows can handle the osi_loop export very well. The shlibtest.c code needs to use both dllimport and dllexport.

burgerrg:
I ran the automated tests on Windows 10, Ubuntu 18.04.3, and macOS 10.14.6. Travis should catch up later today.

owaddell:
It looks like we might be able to call `osi_get_callbacks(0)` where we had the other calls to `uv_run` (in `osi_exit` and `swish_service`).

Chris pointed out that you may not want to do that in case the code for the timeout == 0 case ever changes, so another option might be to consolidate the common code in an `osi_close_loop` helper called in those places.

burgerrg:
I like the idea of calling osi_get_callbacks(0) when we're closing the loop better than a new helper function. Thanks!

burgerrg:
I made the changes, and the tests all pass.

PR #75, opened 2019-10-29
SQLite and LibUV Updates

laqrix:
External library updates

### DIFF
--- a/src/swish/io.ms
+++ b/src/swish/io.ms
@@ -179,32 +179,33 @@
         primes))
      '()
      buffer-sizes))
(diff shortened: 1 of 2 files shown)


burgerrg:
The io.ms tests occasionally have trouble on macOS, so I made the test more tolerant of macOS file-system events for new files.

burgerrg:
The test is still failing on macOS but doesn't exhibit that behavior on my laptop. I'll look into it some more.

laqrix:
Thanks for the review Bob.

PR #76, opened 2019-10-31
Improve debugging experience

owaddell:
This pull request uses native exceptions to preserve more debugging context. While it includes several breaking changes, most code can be adapted to these changes with minimal editing and folks can gradually opt in to the new features from there.

Code that uses the older `(catch e0 e1 ...)` form with `match` patterns of the form `#(EXIT ,reason)` is compatible with these changes for now. At some point, these forms may be deprecated; users should update code to use the new forms described below before then.

The core change here is opt-in. It adds the following forms to help preserve debugging context:
* `(try e0 e1 ...)` as an alternative to the existing `(catch e0 e1 ...)`
* `(throw reason)` and `(throw reason inner)` as alternatives to `(raise reason)`, and
* a match extension `` `(catch ,r)`` that matches exceptions trapped by `(try e0 e1 ...)` or the existing `(catch e0 e1 ...)`
* a match extension `` `(catch ,r ,e)`` that also matches exceptions trapped by `(try e0 e1 ...)` or the existing `(catch e0 e1 ...)` and may provide debugging context via `,e`

To adapt existing code to the breaking changes:
* Replace `match` patterns of the form `#(DOWN ,m ,pid ,reason)` with the match extension `` `(DOWN ,m ,pid ,reason)``.
* Replace `match` patterns of the form `#(EXIT ,pid ,reason)` with the match extension `` `(EXIT ,pid ,reason)``.

Where `monitor` sent a vector of the form `#(DOWN m pid reason)` and where `kill` or `link` sent a vector of the form `#(EXIT pid reason)` they now send records that may preserve debugging context for `reason`. Use the `DOWN` and `EXIT` match extensions to match these new record types. In practice, this means changing `#` to `` ` `` in a few `match` clauses.

Optionally use the match extensions `` `(DOWN ,m ,pid ,reason ,e)`` and `` `(EXIT ,pid ,reason ,e)`` which may provide debugging context via `,e`.

Other breaking changes affect tooling:
* The two-argument form of `dbg` is removed. Use the more general `ps-fold-left` instead.
* The mat file format now records stacks in JSON format and may include multiple stacks for a single failed test, so `mat-result-stack` is now `mat-result-stacks`. We omit stacks that merely show the test harness calling into user code.
* The `coerce` function now writes conditions in JSON format. The `Log.db3` tables `child_end`, `gen_server_terminating`, and `supervisor_error` now include a `details` column that may contain JSON data that provides the output of `exit-reason->english` along with stack information.

The diagnostics web pages now show debugging information where available.

- [X] Improve test coverage __before__ making any changes
- [X] Make breaking changes to put jelly back inside doughnut
- [X] Try using these changes in internal projects
- [X] Get feedback and iterate
- [x] Update documentation
- [x] Rebase
- [x] Try `try` again


owaddell:
Pushed again because my earlier rebase broke app.ms in a few intermediate commits.

doc/swish/event-mgr.tex
```diff
@@ -232,3 +232,36 @@
...
+than \code{normal} or \code{shutdown}.
+Otherwise it returns \code{\#t} if \var{x} is
+not \code{normal} or \code{shutdown}.
+
+\defineentry{normalize-exit-reason}
+\begin{procedure}
+  \code{(normalize-exit-reason \var{r} \opt{\var{e}})}
+\end{procedure}
+\returns{} \var{reason} and \var{details}
+
```
burgerrg: What happens when \var{e} is not specified?

src/swish/supervisor.ss
```diff
@@ -352,23 +354,27 @@
...
+             (values reason err)])])]
+      [#(error ,reason)
+       (values reason #f)]))
 
   (define (monitor-child pid)
     (monitor pid)
     (unlink pid)
     (receive (after 0 'ok)
-      [#(EXIT ,@pid ,reason)
+      [`(EXIT ,@pid ,reason ,err)
```
burgerrg: `err` does not appear to be used. Same for `,err` below.

doc/swish/erlang.tex
```diff
@@ -1064,10 +1156,44 @@ \subsection{Miscellaneous}
...
-(initially \var{base}). It returns the final accumulator value.
+% ----------------------------------------------------------------------------
+\defineentry{ps-fold-left}
+\begin{procedure}
+  \code{(ps-fold-left \var{id<?} \var{base} \var{f})}
+\end{procedure}
+\returns{} see below
+
+The \code{ps-fold-left} procedure folds over the process table
+ordered by \var{id<?} on \code{process-id} and calls \var{f} on
```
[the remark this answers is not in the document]
owaddell: Thanks for catching this.

doc/swish/event-mgr.tex
owaddell: Good catch. I've added some text that may help.

src/swish/db.ss
```diff
@@ -164,20 +164,22 @@
...
-    (sqlite:close ($state db))
-    'ok)
+    (let ([x (catch (flush state))])
+      (vector-for-each (lambda (e) (sqlite:finalize (entry-stmt e)))
+        (hashtable-values (cache-ht ($state cache))))
+      (finalize-lazy-statements ($state cache))
+      (sqlite:close ($state db))
+      (match x
+        [#(EXIT ,reason) (raise reason)]
+        [,_ 'ok])))
```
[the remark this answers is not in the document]
owaddell: Ah, right. My WIP commit message was really unclear. I had left it as `catch` and `#(EXIT ...)` with the intent of squashing it into the "fix: bad input to db:log could hang db:stop indefinitely" commit. I've added the missing commit that flips these to the new style to be squashed in as part of the "retain more debugging context" commit. Hopefully this will all make sense in the final rebase.

src/swish/supervisor.ss
owaddell: Good point. I've updated it to use the value and added some tests.

owaddell:
Thanks for the quick review, Bob! I've taken a crack at making the `pps` mat more robust. Curious to see what the build server turns up. I also tweaked the `swish-test` HTML output to replace the tedious hover mechanics for viewing stacks with more sensible toggle behavior.

Hmm, somehow I seem to have started a review of my own. Not sure how that happened.

owaddell:
Minor tweaks during the rebase:
* "test: fix mats that booted using the default log-file"
   reverted unnecessary change of quote to backquote to minimize diffs
* "fix: update execute-with-retry-on-busy to look for current error codes"
  removed start-silent-event-mgr from "errors" mat since we now start a silent event mgr for all of db.ms
* "retain more debugging context"
  reword commit message to add missing "without" before "attributing"
* "breaking change: log fault-condition in gen-server and supervisor tabls"
  reword commit message to add missing ">" to "<gen-server-terminating"

owaddell:
Thanks for all the review work on this one, Bob!

I just realized that I didn't document what `stack->json` does if the _max-depth_ argument is omitted. Folks here have asked whether the default max depth can be changed, so I may ponder that before pulling the trigger.

owaddell:
Bob, thanks for your quick feedback and patience with continued updates to this branch! I hope to iron out the wrinkles so we can consolidate the breaking changes in this pull request.

Is there anything fundamentally misguided in the `limit-stack` experiment? I may have left enough bread crumbs in the commit log to convey what I was going after. It looks like it adds 544 bytes to the stack on a6le, which may be worth it to simplify the debugging information we log. I suspect cp0 and wpo optimization might thwart this approach by unravelling `$limit-stack`, so it might be merely advisory to `walk-stack`.

burgerrg:
I like the limit-stack form and would find it useful. 544 bytes (68 words) seems fine.

owaddell:
@burgerrg I moved some commits from the isolate-mat pull request into this one since some of them need to be squashed in here. I also squashed the WIP commits you had already approved down where they belong. I left the `limit-stack` changes spread out across a few WIP commits so we have options. I fixed a few more bugs along the way. Hope this doesn't create too much confusion.

owaddell:
Not sure what caused the problem on the ta6osx build since the isolate-mat branch built successfully.

burgerrg:
The ta6osx build complained about files being out of date, so I think it tickled the HFS+ timestamp resolution limitation. I restarted the job, and it failed for swish-build because it timed out, probably because the host machine was too host. The next run passed.

owaddell:
Final (?) rebase before merging this to master. Minor changes since the point where it was last approved:
- parameterize `print-level` and `print-length` in `dump-process-table` which is called by `pps` and `dbg`
- moved the commit "fix: release process-parameters when process dies" back over to #77 where it was originally since that keeps the inspection history intact
- used \hyperlink to point to `walk-stack` in documentation of `dump-stack` and `stack->json` (\hyperref kept pointing us back to the enclosing section)
- fixed bug, added test, and updated documentation for case where the `walk-stack-max-depth` change meant we could now call the default `truncated` procedure in `walk-stack`
- added missing bad-arg tests for `walk-stack-max-depth` and `with-process-details`
- added missing tests for `walk-stack-max-depth` effect on `walk-stack` and `dump-stack` default cases
- squashed the different flavors of `limit-stack` keeping the one last reviewed which attempts to embed source from call site
- added missing test for multiple-value return through `limit-stack`

PR #77, opened 2019-12-05
Update isolate-mat and fix process-parameters

owaddell:
folks need ways to tune isolate-mat

make sure we don't hold onto process parameter values when process dies


src/swish/testing.ss
```diff
@@ -135,34 +136,64 @@
...
+                       [deadline (lookup 'process-cleanup-deadline)]
+                       [kill-delay (lookup 'process-kill-delay)])
+           #`(add-mat 'name 'tags
+               #,(replace-source x ;; use isolate-mat source in timeout continuation
+                   #'(lambda ()
+                       ($isolate-mat timeout-ms deadline kill-delay
+                         (lambda () e1 e2 ... (void)))
+                       (void))))))]
+      [(_ name tag-list e1 e2 ...)
+       #'(isolate-mat name (isolate-mat-settings [tags tag-list]) e1 e2 ...)]))
```
burgerrg: Will this case also preserve the source information?

owaddell: Excellent question. In my testing the expected source information is preserved in both cases, though it's not immediately obvious why it works.

I half expected we might need to replace `_` on the original line 195 with a pattern variable and reference than in place of `isolate-mat` on 196. That may be more reliable or at least easier to fathom.

Another option is to use `#'name` instead of `x` as the source for `replace-source`, but that may mislead someone looking at the stack to think that mat names are bound to procedures.

owaddell:
Looks like Travis had trouble with the "choco install" stuff in a couple of the Windows builds earlier. Maybe the rebase and rebuild will jog it loose.

owaddell:
Final (?) rebase. Took the suggestion of shortening `isolate-mat-settings` to `settings`. Moved the "fix: release process-parameters when process dies" commit back to this pull request from #76 where it had been moved by mistake.

owaddell:
Not sure how to nudge Travis to rebuild short of re-pushing. Looks like timeouts and some issues setting up the Windows build environment.

owaddell:
Repush to nudge Travis.

PR #79, opened 2020-01-07
Added pregexp-replace/transform*

wintonpc:
`pregexp-replace/transform*` allows the replacement to be a transformation of the substring being replaced.

I didn't add to swish.pdf since none of the other pregexp procedures are documented there.

### DIFF
--- a/src/swish/pregexp.ms
+++ b/src/swish/pregexp.ms
@@ -78,6 +78,13 @@
   (pregexp-replace* "te" "liberte egalite fraternite" "ty")
   "liberty egality fratyrnity"
 
+  (pregexp-replace/transform* "te" "liberte egalite fraternite" string-upcase)
+  "liberTE egaliTE fraTErniTE"
+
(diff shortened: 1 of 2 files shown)


owaddell:
Thanks for the pull request @wintonpc . Bob ported the pregexp library to Swish and is more familiar with the code, so I've asked him to review the changes.

I haven't grokked what `pregexp-replace-aux` is doing, but did notice that it handles a case where `#f` is among the `backrefs` passed in. The new code doesn't yet handle that case when it constructs `ins`. For example, try:
```
(pregexp-replace/transform* "(a)|(b)" "b" (lambda x "foo"))
```

For documentation, maybe we should add a "Regular Expressions" subsection after the "String Utilities" subsection in erlang.tex, move the "For regular expression support ... `\cite{pregexp}`." and the \defineentry for `re` to that subsection, and add a new entry for `pregexp-replace/transform*`.

laqrix:
I accidentally closed this pull request, however, it appears to be stale. @wintonpc if you are still interested in this change, please reopen this or resubmit a new request.

PR #80, opened 2020-01-27
fix test broken by Chez Scheme 9.5.3 library-manager updates

owaddell:


burgerrg:

PR #81, opened 2020-01-27
SQLite 3.32.1

laqrix:
Creating a draft pull request for the newest SQLite. We should let the release simmer a bit before merging this.

### DIFF
--- a/src/swish/sqlite3.h
+++ b/src/swish/sqlite3.h
@@ -123,9 +123,9 @@ extern "C" {
 ** [sqlite3_libversion_number()], [sqlite3_sourceid()],
 ** [sqlite_version()] and [sqlite_source_id()].
 */
-#define SQLITE_VERSION        "3.30.1"
-#define SQLITE_VERSION_NUMBER 3030001
-#define SQLITE_SOURCE_ID      "2019-10-10 20:19:45 18db032d058f1436ce3dea84081f4ee5a0f2259ad97301d43c426bc7f3df1b0b"
+#define SQLITE_VERSION        "3.32.1"
+#define SQLITE_VERSION_NUMBER 3032001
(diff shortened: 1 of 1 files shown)


burgerrg:
Good call letting it simmer. The point release came soon thereafter!

laqrix:
I noodled with Generated Columns. I got it to work except that usage of 'localtime' is considered non-deterministic and thus fails. That's a bummer. I also discovered that the tool I use for cracking open DB3 files doesn't support this version yet, which means it can't even load a SQLite file with generated columns in it. I think we should hold off a bit longer until there is better support from viewing tools.

burgerrg:
SQLite Expert now uses SQLite 3.31.1. Do you have any other tool that doesn't yet support it?

laqrix:
I've been using DB Browser for SQLite for a while now. It looks like they released an Alpha version on 4/15 that might add support.

laqrix:
I updated the branch but DB Browser for SQLite still cannot open a database once a generated column is added. I've been tracking this issue https://github.com/sqlitebrowser/sqlitebrowser/issues/2107 for updates.

laqrix:
I had convinced myself that we should go ahead and absorb the latest SQLite even if other tools don't support the new features. Then I discovered that SQLite 3.32.1 had been released. I've updated to the latest and rebased on master. I think we should proceed forward with this version.

laqrix:
Thanks for the quick review Bob!

justinclift:
As a data point, SQLite 3.32.2 was released recently to fix an important bug:

&nbsp; &nbsp; https://sqlite.org/releaselog/3_32_2.html

Might be useful to bump the included version up again. :smile:

laqrix:
@justinclift Thanks for the information. It appears I am no longer receiving email updates about SQLite releases. I've created a separate pull request for 3.32.2.

PR #82, opened 2020-01-28
minor testing fixes

owaddell:
- flush mat output give better odds of parsing test results on failure
- bump timeout in application-shutdown test that seems to upset ta6osx

### DIFF
--- a/src/swish/app.ms
+++ b/src/swish/app.ms
@@ -412,7 +412,8 @@
             [locking-up 'ok])
           (printf "armed and ready\n")
           (send (whereis 'bad-sup) 'disarm)
-          (receive (after 1000 (osi_exit 4)))))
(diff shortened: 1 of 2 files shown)


src/swish/app.ms
```diff
@@ -396,7 +396,8 @@
               (cond
                [(whereis 'profiler) =>
                 (lambda (pid)
-                  (gen-server:call pid 'save 'infinity))]))
+                  (time
```
burgerrg: You probably don't want to leave the time call in here?

owaddell: Oops! Thanks for catching this.

owaddell:
Thanks, Bob! I wish there were a better way to deal with the vagaries of the Traviss-Mac-6.local build host.

PR #84, opened 2020-02-12
fix: unlock callback and buffer for outstanding read when closing port

laqrix:
We typically use a pattern of an async reader process. If that process
is blocked on IO while another process closes the underlying port, the
async reader would not only never return, but would also never be collected.

The osi_read_port callback closed over the process and is locked while
the read is outstanding. Once closed, LibUV will never call its callback
routine which unlocks the Scheme callback and buffer.

The fix unlocks both the Scheme read callback and the read buffer and calls
the read callback with UV_EOF. This allows the Scheme code to process
an end-of-file and any remaining computation including winders.

The order of the read callback and the close callback should not matter.
close-osi-port immediately sets the port-handle to #f, which makes any
subsequent read calls on alternate processes safely fail.

### DIFF
--- a/src/swish/osi.c
+++ b/src/swish/osi.c
@@ -466,9 +466,19 @@ static ptr write_stream_port(uptr port, ptr buffer, size_t start_index, uint32_t
 }
 
 static void close_stream_cb(uv_handle_t* handle) {
+  // This code does not check and unlock the write_callback and
+  // write_buffer because LibUV explicitly cancels write operations
+  // with UV_ECANCELED before processing this callback.
   stream_port_t* p = container_of(handle, stream_port_t, h.stream);
   ptr callback = p->close_callback;
+  ptr read_callback = p->read_callback;
+  ptr read_buffer = p->read_buffer;
   free(p);
+  if (read_callback) {
+    Sunlock_object(read_buffer);
+    Sunlock_object(read_callback);
+    osi_add_callback1(read_callback, Sfixnum(UV_EOF));
+  }
   if (callback) {
     Sunlock_object(callback);
     osi_add_callback1(callback, Sfixnum(0));
--- a/src/swish/io.ms
+++ b/src/swish/io.ms
@@ -750,4 +750,78 @@
     [,@expected (format "~s" l)])
    'ok))
 
+(isolate-mat read-leak ()
+  (let* ([hostname "::1"]
+         [listener (listen-tcp hostname 0 self)]
+         [test-port (listener-port-number listener)]
+         [gbuff (make-guardian)]
+         [gproc (make-guardian)])
+    (on-exit (close-tcp-listener listener)
+      (let-values ([(cip cop) (connect-tcp hostname test-port)])
+        (on-exit (begin (close-input-port cip) (force-close-output-port cop))
+          (receive (after 5000 (throw 'timeout-connecting-tcp))
+            [#(accept-tcp ,@listener ,sip ,sop)
+             (on-exit (force-close-output-port sop)
+               (let* ([me self]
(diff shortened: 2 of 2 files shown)


burgerrg:
@laqrix, I pushed a commit to fix the automated tests.

laqrix:
I believe the case for write_buffer and write_callback is less necessary.  If you are using Scheme's port abstraction, close-port will induce a flush-output-port. If there is a write outstanding, you are likely to have a multi-threaded race problem. If you are in the same process, the write will fully complete before the close.  If we are concerned someone is bypassing the Scheme port abstraction, then I completely agree, but we will need to find a way to test this case at the OSI level.

laqrix:
I changed my mind. We should update write_callback and write_buffer for completeness and correctness. I'm not sure how to handle the testing, but we should try something.

burgerrg:
Yes, I think it's a good idea to keep it complete & consistent. For testing, try a pipe to another Swish process that doesn't read its input. Fill the pipe to get a write that doesn't complete. Then close the port and watch the sparks fly!

laqrix:
Take a look at the `write-leak` mat. Perhaps I missed an obvious path. When I run it, I always get UV_ECANCELED from the write_callback before the close callback fires (at least on Windows and Linux). That matches my expectations at an OS level. Any suggestions to improve the test?

laqrix:
@owaddell found the documentation for libuv's [uv_close](http://docs.libuv.org/en/v1.x/handle.html) explicitly states:

> In-progress requests, like uv_connect_t or uv_write_t, are cancelled and have their callbacks called asynchronously with status=UV_ECANCELED.

burgerrg:
Since uv_close explicitly guarantees that outstanding write callbacks are invoked before the descriptor is closed, should we simply leave the code as is?

burgerrg:
I'm a little surprised that libuv doesn't handle outstanding reads the same way.

laqrix:
I was also surprised. I think the asymmetry in the code is unfortunate, but reasonable given the circumstance. Maybe we should add the write-leak test as a regression to make sure the behavior never changes in the future?

burgerrg:
Yes, I think it's fine to leave the code asymmetric and to add a write test to verify proper write callback usage when a handle is closed. Thanks!

laqrix:
@burgerrg I added a comment to `close_stream_cb` explaining the LibUV behavior (and updated the commit comment). I renamed the tests to `read-leak` and `write-leak`. My second force push also included a check that the winders ran in the write-leak test.

src/swish/osi.c
```diff
@@ -466,9 +466,19 @@ static ptr write_stream_port(uptr port, ptr buffer, size_t start_index, uint32_t
...
+  // with UV_ECANCELED before processing this callback.
   stream_port_t* p = container_of(handle, stream_port_t, h.stream);
   ptr callback = p->close_callback;
+  ptr read_callback = p->read_callback;
+  ptr read_buffer = p->read_buffer;
   free(p);
+  if (read_callback) {
+    Sunlock_object(read_buffer);
+    Sunlock_object(read_callback);
+    osi_add_callback1(read_callback, Sfixnum(UV_EOF));
```
burgerrg: Should we use UV_ECANCELED for consistency?

laqrix: I think that might create unexpected behavior from the Scheme port perspective. Where you might have received either a values or an eof-object, a cancel will cause an exception. At least in the cases I've been working through, my primary process has explicitly closed the port, and I expect an eof-object, not an exception.

burgerrg: That seems reasonable.

laqrix:
Thanks for the review!

PR #85, opened 2020-02-26
Breaking: Improve HTTP flexibility

laqrix:
These changes are to improve flexibility, maintainability, and security of the HTTP subsystem. These changes force the application to add the desired HTTP servers to the supervision tree.

Here is the summary of the breaking changes:
* added HTTP request timeout
* disabled automatic form handling
* represent HTTP header and params as JSON for better logging and querying
* removed `http-port-number`
* removed `web-dir`
* disable dynamic pages by default

These changes enable:
* serving files directly from code
* multiple HTTP servers
* serving files from multiple disjoint paths

To port existing code, in place of `(http-port-number 54321)`, the following code will enable the same functionality:
``` scheme
(http:add-file-server 'http 54321 (path-combine (base-dir) "web")
  (http:options
   [file-search-extensions '(".ss" ".html")]
   [file-transform http:enable-dynamic-pages]))
```


doc/swish/http.tex
```diff
@@ -273,10 +416,11 @@
 \returns{} a string \alt{} \code{\#f}
 
 The \code{http:find-param} procedure returns the value associated with
-\var{name} in \var{params} if present and \code{\#f}
-otherwise. Parameter comparisons are case-sensitive. If \var{name} is
-not a string, exception \code{\#(bad-arg http:find-param \var{name})}
-is raised.
+\var{name} in \var{params} if present and \code{\#f} otherwise.  If
+\var{name} is a symbol, the lookup uses \code{eq?}.  If \var{name} is
+a string, it is first mapped to a lower-case symbol before performing
```
burgerrg: http:find-param does a case-sensitive comparison, so the string isn't downcased.

doc/swish/http.tex
```diff
@@ -286,28 +430,30 @@
 
 The \code{http:get-param} procedure returns the value associated with
 \var{name} in \var{params} if present and raises exception
-\code{\#(invalid-param \var{name} \var{params})} otherwise. Parameter
-comparisons are case-sensitive. If \var{name} is not a string,
-exception \code{\#(bad-arg http:get-param \var{name})} is raised.
+\code{\#(http-invalid-param \var{name} \var{params})} otherwise.  If
+\var{name} is a symbol, the lookup uses \code{eq?}.  If \var{name} is
+a string, it is first mapped to a lower-case symbol before performing
```
burgerrg: http:get-param does a case-sensitive comparison, so the string isn't downcased.

doc/swish/http.tex
```diff
@@ -261,9 +404,9 @@
 \returns{} an unsigned integer \alt{} \code{\#f}
 
 The \code{http:get-content-length} procedure returns the unsigned
-integer value associated with \code{Content-Length} in \var{header} if
+integer value associated with \code{"Content-Length"} in \var{header} if
```
burgerrg: \code{content-length}

doc/swish/http.tex
```diff
@@ -333,18 +479,18 @@
...
 \begin{procedure}
   \code{(http:write-header \var{op} \var{header})}
 \end{procedure}
 \returns{} unspecified
 
-The \code{http:write-header} procedure writes the HTTP \var{header},
+The \code{http:write-header} procedure writes the HTTP \var{header}
 and trailing CR LF to the binary output port \var{op}.
 
 \var{header} is an association list. If \var{header}'s keys are not
```
burgerrg: http:read-header returns a JSON object, but http:write-header takes an association list. The association list may be easier to express when writing, and the JSON is faster to query when reading.

burgerrg: For consistency, consider making header a JSON object. In http:respond, note that it side-effects the JSON header by adding the keys listed in the description if they're not already present.

laqrix:
I had considered switching http:write-header. My original concerns where that 1) HTTP headers are allowed to be duplicate (though I would not encourage this behavior and to work around you can make a single value a comma-separated list), and 2) the potential copying to avoid side effects.  If we're ok with documenting the side effecting, then we could try it.

My new concerns would be 1) a user needs to always down-case their JSON keys or we may accidentally write duplicate HTTP header fields (in particular the ones we set! into the user's object) and 2) the impact on the mats may be large -- any mats that observe the header written by http:write-header is probably not written to tolerate arbitrary ordering.  Would we continue to support association list input, or support JSON-only?

burgerrg:
I think the advantages of alist headers outweigh the benefits. I propose changing the documentation to use *header-alist* for clarity. I'll make the change on the WIP commit.

burgerrg:
I suggested the http:add-server and http:add-file-server macro changes so that we get procedure arity checking for free.

laqrix:
You're doing great Bob! I found a bug in http:make-default-url-handler where it ignored the return value of the handler -- a particularly ugly bug if your handler wants to use http:switch-protocol.  I also added a separate commit (I would rather not squash) relating to our usage of Content-Length in http:respond.

burgerrg:
Good fixes!

burgerrg:
Do you want to use arg-check in http:respond and http:respond-file?

laqrix:
Thanks for the updates Bob.  I didn't put arg-check in http:respond and http:respond-file because http:write-status and http:write-header both do checks.  In a failure, it may give a misleading "who", but I thought that was better than double checking.

laqrix:
Thanks again for the review Bob!

PR #86, opened 2020-03-11
add WebSocket support

laqrix:


src/swish/http.ms
```diff
@@ -1409,24 +1409,27 @@
...
+       "Host: 127.0.0.1"
+       "Upgrade: websocket"
+       "Sec-Websocket-Version: 13"
+       "Sec-Websocket-Key: dGhlIHNhbXBsZSBub25jZQ=="))
+    (put-bytevector op (string->utf8 "\r\n"))
+    (flush-output-port op)
+    (let ([s (receive [#(ws:init ,s) s])])
+      (receive (after 20 'ok))
+      (receive (after 1000 (throw 'no-server-close))
+        [#(ws:closed ,@s 1002 "no response from ping") 'ok]))))
```
burgerrg: You might add a timeout clause to the receive for ws:init, and I don't think you need the 20 ms wait any more.

burgerrg:
```
(receive (after 1000 (throw 'no-ws:init))
 [#(ws:init ,s)
  (receive (after 1000 (throw 'no-server-close))
   [#(ws:closed ,@s 1002 "no response from ping") 'ok])])
```

laqrix: You're right. I'm trying a different philosophy on timeouts. For the ws:init case it feels like an assertion for the setup of the whole test -- I'm letting isolate-mat's timeout handle that. For the later timeout, I'm checking behavior so I wanted a specific throw.

I'll remove the 20ms wait. Thanks.

laqrix: I changed my mind. That looks nice without the let binding. I'll take that change too.

laqrix:
Thanks again Bob and Oscar for the review!

PR #89, opened 2020-04-07
fix: an error in osi_get_ip_address inside a callback crashes the event-loop

laqrix:
The direct call to osi_get_ip_address inside the callback procedures
for connect-tcp and listen-tcp can fail. When it does, the event-loop
crashes the entire Swish process.

Failing to get the ip address for the connection does not invalidate
the connection. We try to use a reasonable name if available.

### DIFF
--- a/src/swish/io.ss
+++ b/src/swish/io.ss
@@ -892,6 +892,12 @@
 
   (define (port-number? x) (and (fixnum? x) (fx<= 0 x 65535)))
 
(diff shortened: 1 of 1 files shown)


laqrix:
In connect-tcp, we were already using format to generate the name for wait-for-io (non-error case).

burgerrg:
> In connect-tcp, we were already using format to generate the name for wait-for-io (non-error case).

Ah, yes, thank you!

PR #91, opened 2020-05-09
minor documentation updates

owaddell:
- fixed a couple return types
- wondering about return value of `watcher:shutdown-children`
- took a stab at getting return type on same page as makedecl header box


burgerrg:
I think it's fine for watcher:shutdown-children to return unspecified and to update the document to say so.

owaddell:

PR #92, opened 2020-05-20
add spawn-os-process-detached

laqrix:
This change adds `spawn-os-process-detached` for cases when we want to fire-and-forget and external process. This is useful when starting processes in user-space that do not require the overhead of being a service.

Testing is light because the external process is detached and thus hard to monitor. There are two cases of particular interest that can be done manually. 1) Fire up and external process, then exit Swish. The external process should remain. 2) Fire up an external process, then have it exit. Swish should remain and not crash (which would happen if we included invalid callback pointers in the osi process struct).

### DIFF
--- a/src/swish/io.ss
+++ b/src/swish/io.ss
@@ -91,6 +91,7 @@
    signal-handler
    signal-handler-count
    spawn-os-process
+   spawn-os-process-detached
    stat-directory?
    stat-regular-file?
    tcp-listener-count
@@ -852,6 +853,13 @@
            (make-iport name (@make-osi-port name from-stderr) #t))
          os-pid)]
        [(,who . ,errno) (io-error path who errno)])))
+
+  (define (spawn-os-process-detached path args)
+    (unless (and (list? args) (for-all string? args))
+      (bad-arg 'spawn-os-process-detached args))
+    (match (osi_spawn_detached* path args)
+      [(,who . ,errno) (io-error path who errno)]
+      [,os-pid os-pid]))
 
   ;; TCP/IP Ports
 
--- a/src/swish/osi.c
+++ b/src/swish/osi.c
@@ -617,9 +617,11 @@ static void free_argv(char** argv) {
 
 static void process_exit_cb(uv_process_t* process, int64_t exit_status, int term_signal) {
   ptr callback = (ptr)process->data;
-  osi_add_callback3(callback, Sinteger32(process->pid), Sinteger64(exit_status), Sinteger32(term_signal));
-  Sunlock_object(callback);
-  process->data = 0;
+  if (callback) {
+    osi_add_callback3(callback, Sinteger32(process->pid), Sinteger64(exit_status), Sinteger32(term_signal));
+    Sunlock_object(callback);
+    process->data = 0;
+  }
   uv_close((uv_handle_t*)process, close_handle_data_cb);
 }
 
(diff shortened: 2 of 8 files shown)


src/swish/io.ss
```diff
@@ -853,6 +854,14 @@
          os-pid)]
        [(,who . ,errno) (io-error path who errno)])))
 
+  (define (spawn-os-process-detached path args)
+    (unless (and (list? args) (for-all string? args))
+      (bad-arg 'spawn-os-process-detached args))
+    (with-interrupts-disabled
```
burgerrg, line 860: There's no need to disable interrupts here because there are no interruptable parts inside.

src/swish/osi.c
```diff
@@ -888,6 +888,54 @@ ptr osi_spawn(const char* path, ptr args, ptr callback) {
...
+    .gid = 0
+  };
+  p->data = 0;
+  int rc = uv_spawn(osi_loop, p, &options);
+  free_argv(argv);
+  if (rc < 0) {
+    uv_close((uv_handle_t*)p, NULL);
+    return osi_make_error_pair("uv_spawn", rc);
+  }
+  return Sinteger32(p->pid);
```
burgerrg, line 936: This leaks the uv_process structure and doesn't close it. We need to define an exit_cb that mimics process_exit_cb but without the callback.

src/swish/io.ss
laqrix: Good catch!

src/swish/osi.c
laqrix: Upon further reading, I don't think we want to hook an exit_cb, as that is fired when the process exits. By definition, we don't want to be monitoring the process. I do however see in the LibUV docs that I should've called `uv_unref`. Maybe I can call `uv_close` directly here. See what you think of the patch.

burgerrg: We still need to free the memory for p.

burgerrg: I tried to use an automatic variable for the uv_process_t and avoid the malloc & free altogether. Unfortunately, uv_close threw an invalid memory reference when I did.

burgerrg: I pushed commits to simplify the closing process, and it appears to work on the Mac. I'll try it on Windows shortly.

burgerrg: On Unix systems, the child is still watched even after we uv_close. I'm going to propose an exit_cb because it cleans everything up correctly.

burgerrg: Even with the latest commit, I get an invalid memory reference when I hit CTRL-C on the Mac from a loop that continuously spawns "sync". Something's still fishy here.

burgerrg: I found that uv_close always asynchronously calls its callback, even when it's NULL. This means we shouldn't call it with NULL and then immediately free the data structure. See the latest commit.

burgerrg:
To summarize, libuv always watches the processes it spawns. On macOS, even if we uv_close the process handle, the watched process remains even after it exits as seen by `ps -x`. As a result, the safest option for all platforms is to hook the exit_cb to call close to free the memory.

laqrix:
I tried 2 more alternatives and they both failed. I squashed everything together and will let Travis have another chance to build it.

laqrix:
Thanks Bob!

PR #93, opened 2020-05-21
doc: use imakeidx to eliminate warnings from obsolete index package

burgerrg:


burgerrg:
@owaddell, recent TeXLive distributions have a [known issue](https://github.com/latex3/latex2e/issues/299) that causes a warning:

```
LaTeX Warning: Font shape declaration has incorrect series value `mc'.
```

burgerrg:
I switched from `kpsewhere` to `kpsewhich` because the former is not part of Cygwin's texlive distribution any more.

burgerrg:
I've tested this on Mac with MacTeX, Ubuntu 18 with texlive, and Windows 10 with Cygwin.

burgerrg:
Thanks, Oscar!

PR #94, opened 2020-05-22
libuv v1.38.0

burgerrg:
Libuv now uses cmake for configuration, which eliminates the need for Python and simplifies the build process.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -18,7 +18,7 @@ font.conf:
 	@echo '<?xml version="1.0"?>' > $@
 	@echo '<!DOCTYPE fontconfig SYSTEM "fonts.dtd">' >> $@
 	@echo '<fontconfig>' >> $@
-	@echo "  <dir>$$(dirname $$(kpsewhere lmroman10-regular.otf))</dir>" >> $@
+	@echo "  <dir>$$(dirname $$(kpsewhich lmroman10-regular.otf))</dir>" >> $@
 	@echo '</fontconfig>' >> $@
 
 swish :=\
--- a/doc/run-latex
+++ b/doc/run-latex
@@ -71,11 +71,6 @@ alert()
 
 update_indices()
 {
-  if grep -F -asq "$1.idx" "$1.log"
-  then
-    ${MODE} makeindex -q "$1.idx" -o "$1.ind"
-    changed=yes
-  fi
   if grep -F -asq "$1.thm" "$1.log"
   then
     changed=yes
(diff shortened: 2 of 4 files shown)


burgerrg:
One breaking change on Windows: the name of the libuv DLL is now uv.dll, not libuv.dll.

configure
```diff
@@ -388,22 +361,17 @@ SWISH_LIBPATH="${BASE}/build/release/bin"
...
+    UvLib=libuv/build
     ;;
 esac
-LIBUV_LIBPATH="${BASE}/${UvLib}"
 if [ "${Windows}" = "yes" ]; then
     SCHEME="$(cygpath -au "${SCHEME}")"
     SWISH_INCLUDE="$(cygpath -aw "${SWISH_INCLUDE}")"
     SWISH_LIBPATH="$(cygpath -aw "${SWISH_LIBPATH}")"
     LIBUV_INCLUDE="$(cygpath -aw "${LIBUV_INCLUDE}")"
-    LIBUV_LIBPATH="$(cygpath -aw "${LIBUV_LIBPATH}")"
```
laqrix, line 406: I think this was a little too aggressive deletion. If I have code that needs uv.lib, we need to know where to find it. (LIBUV_LIBPATH)

src/swish/Mf-a6le
```diff
@@ -8,28 +8,28 @@ ifneq (,$(shell "${CC}" --help=warnings 2> /dev/null | grep return-local-addr))
   HUSH:=${HUSH} -Wno-return-local-addr
 endif
 C = ${CC} -m64 -msse2 -fPIC -Wall -Wextra -Werror -O2 ${CPPFLAGS} ${CFLAGS} ${LDFLAGS}
-OsiObj=osi.o sha1.o sqlite.o sqlite3.o
+OsiObj=osi.o sha1.o sqlite.o sqlite3.o ${UvLib}/libuv_a.a
```
laqrix, line 11: This seems like a good move, but slightly changes the meaning of OsiObj across different Mf-a6* files.

burgerrg: I pushed a commit that puts it back. Thanks!

owaddell:
Thanks for sorting this out, Bob! Tested on Fedora 32 (64-bit) and Ubuntu 20.04. The libuv build is really quick now, which is nice. We may have been running their automated tests before, so maybe that's why it seems quicker.

burgerrg:
I think the former build built all the testing libraries. We select just the one we need now. Thanks for the review!

PR #96, opened 2020-06-11
SQLite 3.32.2

laqrix:


laqrix:
Thanks for the quick review Bob!

PR #97, opened 2020-06-22
update README.md with clearer link to Design document

laqrix:
There has been some confusion from consumers (and me) about a header that is also a link.


laqrix:
Thanks for the review Bob. I down cased "Design" in my commit message and merged it.

PR #100, opened 2020-06-23
Run tests in the order specified on command line

owaddell:
Should we mark this as a breaking change? The original behavior wasn't well documented, though there was a mat checking for sorted order of test results.

Does the mat result format need any changes to better handle duplication?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -74,6 +74,11 @@ jobs:
       before_script:
         - git config core.autocrlf false; rm .git/index; git reset --hard
         - choco install make -y
+    - env: TARGET_MACHINE=a6nt
+      os: windows
+      before_script:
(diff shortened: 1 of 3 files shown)


owaddell:
Oops. Looks like I should label this a broken change. Fortunately it looks like it's breaking the same way on i3le, so I should be able to investigate later.

burgerrg:
Thanks for fixing that!

burgerrg:
It's also fine with me if you want to mark this as a breaking change.

The mat result format seems to have adequate information in it.

owaddell:
Re-pushing to see if a6osx build will be happier. Guessing Travis was just loaded down.

owaddell:
Thanks for the whitespace cleanup and Travis updates, Bob!

PR #101, opened 2020-08-10
add (response-timeout) to http:options

laqrix:
`http:respond` and `http:respond-file` use the default 5 second gen-server
timeout which is not enough for slow networks or large amounts of data.

Adding an optional timeout argument to each of these functions helps,
but does not cover the static file handler buried in the http-cache server.

Adding (response-timeout) allows Swish applications to override this
behavior when configuring the server. Using a default of 60 seconds
should allow a 50 MB file over a 10 Mbit/sec connection, or 500 MB
file over a 100 Mbit/sec connection with overhead.

### DIFF
--- a/doc/swish/http.tex
+++ b/doc/swish/http.tex
@@ -287,6 +287,9 @@ used:
   number of milliseconds to wait to receive an HTTP request after
   initial connection \\
 
+  \code{response-timeout} & 60,000 & a positive fixnum; the maximum
+  number of milliseconds to allow the server to send a response \\
+
   \code{header-limit} & 1,048,576 & a positive fixnum; the maximum
   length in bytes of the HTTP request headers \\
 
@@ -533,7 +536,7 @@ strings. If any of its keys are not strings, exception
 
 \defineentry{http:respond}
 \begin{procedure}
-  \code{(http:respond \var{conn} \var{status} \var{header-alist} \opt{\var{content}})}
+  \code{(http:respond \var{conn} \var{status} \var{header-alist} \opt{\var{content} \opt{\var{timeout}}})}
 \end{procedure}
 \returns{} \code{\#t}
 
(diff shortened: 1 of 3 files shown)


doc/swish/http.tex
```diff
@@ -533,7 +536,7 @@ \subsection{WebSocket Protocol}\label{sec:websocket-protocol}
 
 \defineentry{http:respond}
 \begin{procedure}
-  \code{(http:respond \var{conn} \var{status} \var{header-alist} \opt{\var{content}})}
+  \code{(http:respond \var{conn} \var{status} \var{header-alist} \opt{\var{content}} \opt{\var{timeout}})}
```
burgerrg, line 539: When timeout is specified, content is required, so you might use `\opt{\var{content} \opt{\var{timeout}}}` instead.

burgerrg (commented):
Helpful changes! I'd like to know why the Windows mats are failing, but I don't expect to have time to check for a couple weeks.

doc/swish/http.tex
laqrix: That's a nice change. Thanks.

laqrix:
I didn't develop this update on Windows, so it was a bit of a surprise to me too. I'll dig into it and see what I can find.

laqrix:
It seems that Microsoft has gotten Windows to transfer data on the loopback interface as ridiculously fast rates. I've dialed the data up to 1 Gb, and the timeout down to 100ms and that fails as expected (it takes 200 ms to transfer, so there is some buffer).  However, I believe dump-stack is trying to render the stack (and massive bytevector) to a string, and that is taking long enough (and allocates a ton of RAM) that the mat itself times out. I don't really like the "allocate a 1Gb bytevector", but I don't see another option. I'll keep looking for a solution.

laqrix:
I'm wrong about `dump-stack`. I removed its guts and there is still something unusual happening.

laqrix:
It appears my diagnostics impeded the ability for the test to finish cleanly. I increased the data size to 1Gb and reduced the timeout to 50ms. Without diagnostics, my machine took 100ms to transfer the data, so 50ms should be enough time for the timeout to fire.

laqrix:
Reducing data size to 2^29-1 for 32-bit systems, and timeout to 15ms to give room for timing error.

burgerrg (approved):
Looks good, thank you!

laqrix:
Thanks for the review Bob. Squashing the commits down in this last commit.

PR #104, opened 2020-08-31
Automatically rethrow faults from gen-server:call

laqrix:
A typical pattern with gen-server code is to have the server return an `#(ok ,result) or `#(error ,reason) tuple. The client side regularly pattern matches and raises errors in the caller's context.

``` scheme
(match (gen-server:call who ...)
  [#(error ,reason) (throw reason)]
  [#(ok ,result) result])
```

This change enables a gen-server to provide a Scheme value or a fault via make-fault or (try ...) to a client. The client side of the code is simply:

``` scheme
(gen-server:call who ...)
```

It's tempting to re-evaluate Swish's gen-servers to throw exceptions upon error. The existing code allows programmers to determine if they want to use exceptions or not. I propose that we leave Swish's existing interfaces alone and enable this opt-in behavior for applications built on Swish.


burgerrg (approved):
Thanks for this useful update to gen-server!

PR #106, opened 2020-09-08
Unify process notation across pcb record-writer and log-db coerce

owaddell-beckman:
This pull request aims to unify the process notation used by the `pcb` record-writer and log-db's `coerce` so that references to `#<process xyz>` within `console-event-handler` messages and in log-db error details columns can be associated with log-db rows containing `xyz` in pid columns. Please see the commit log for more details.


burgerrg (approved):
Good work!

owaddell-beckman:
Thanks for the review, Bob! (and thanks to Chris for the pre-reviews!)

PR #108, opened 2020-09-23
update child view and query for global-process-id

owaddell-beckman:
- add an explicit rowid column to the child view, since ordering by pid no longer makes sense
- update errors.ss child query to order by rowid rather than process id

This is a breaking change in the sense that `select * from child` will return more columns than it used to.

### DIFF
--- a/src/swish/log-db.ss
+++ b/src/swish/log-db.ss
@@ -428,7 +428,7 @@
         (set-session-id! (+ (or id 0) 1)))
 
       (execute "drop view if exists child")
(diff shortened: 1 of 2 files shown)


burgerrg (approved):
Good work finding and fixing this, Oscar!

burgerrg:
Automatically triggered Travis-CI builds seem to be broken again. Here's the URL for a manually triggered build of this pull request: https://travis-ci.org/github/burgerrg/swish/builds/729770704

owaddell-beckman:

PR #111, opened 2020-11-09
fix $receive bug

owaddell-beckman:
Keep the interactive Ctrl-C fix, but revert select changes from 2f16c0f2f2c1246193368680682a68cc6dfcae46 that broke $receive.

Co-authored-by: Bob Burger <bburger@indigobio.com>

**Fixes**

This pull request fixes a rare race condition where:
1. `$receive` reaches the end of the inbox and calls `enable-interrupts`
2. the timer interrupt preempts the subsequent call to `yield`
3. `@send` delivers a message to the process that called `$receive` and places that process on the `run-queue`
4. the call to `yield` resumes and removes that process from the run queue

**Proposed changes**

* Leave interrupts disabled until `yield` completes the context switch, as in 2f16c0f2f2c1246193368680682a68cc6dfcae46.
* Introduce `@yield-preserving-interrupts` to clarify the calls in `$receive`.
* Simplify the code by letting `pcb-sic` match the Chez Scheme's internal `disable-count`, since we just care about the relative disable count when switching from one process to the next.


burgerrg (approved):

owaddell-beckman:
Updated the new `keyboard-interrupt-receive` test which failed sometimes, e.g., in `make coverage`. Fixed by setting the `keyboard-interrupt-handler` in the spawned process before sending the final "continue" message.

laqrix:
Thanks Oscar. These changes look good.

burgerrg (approved):

owaddell-beckman:
Bob and Chris, thanks for your help in resolving this!

PR #112, opened 2020-11-10
test: try to make test cache-including test more robust

owaddell-beckman:
**Fixes**

This is an attempt to make the `cache-including` test slightly more robust. This test has failed sometimes on OSX.

**Proposed changes**

* Modify the test to watch for file modifications that we expect the http cache to notice. While there is no guarantee that all processes will see the notification in a timely manner, having the test process wait may improve the odds that the http cache process has gotten the message by the time we make our next request.


owaddell-beckman:
Travis didn't seem to care for this change on OSX. The `cache-including` test now times out.

owaddell-beckman:
macOS builds pass on actual hardware. Not sure about Travis. We may have more luck with Github Actions.

PR #113, opened 2020-11-17
add cmake to Mac requirements

burgerrg:


burgerrg:
@laqrix, I found this when trying out a clean install of macOS Big Sur.

burgerrg:
Since libuv uses cmake, we don't need the full XCode any more. The Command Line Tools are sufficient, and Homebrew automatically installs them.

laqrix (approved):
I think I stumbled into this, but had forgotten that we had this in our readme. Good find!

PR #114, opened 2021-03-05
fix performance bug affecting spawn and spawn&link

owaddell-beckman:
**Fixes**

Version 2.0.1 calls parameter filters when installing the values of inherited parameters in spawned processes. This adds unnecessary overhead to `spawn` and `spawn&link`.  The overhead is particularly steep for the inherited `app:path` parameter, since its filter makes an expensive call to `get-real-path`.

To see this, try the following before and after the patch in this PR:
``` scheme
> (app:path (osi_get_executable_path))
> (time
    (do ([i 0 (#3%fx+ i 1)]) ((#3%fx= i 1000000))
      (spawn values)))
```

**Proposed changes**

`make-inherited-parameter` returns a wrapper that applies the filter, if any, before setting the internal inherited parameter. The internal parameter added to `inherited-parameters` performs no checking, since the values were already filtered when the parameters were set.


burgerrg (approved):
Should we bump the copyright to "2018-21"?

Thanks for making this improvement and fixing the typos!

owaddell-beckman:
I'm happy to bump the swish.tex copyright to 2018-2021 as part of the doc tweaks, if that seems like the right thing to do.
We're currently using a single year 2018 in the coding standards. Do we leave those as is?

burgerrg:
Since we haven't changed the coding standards in a substantial way, 2018 is fine.

owaddell-beckman:
I went with the full year so that it's more likely to turn up the next time we grep.

burgerrg (approved):

owaddell:
Thanks for the timely review, Bob!

PR #115, opened 2021-03-18
extend db integer tests

burgerrg:
I noticed that the 2-bit integer tests in db.ms didn't work as I originally planned. This change fixes it.


owaddell (approved):
Thanks for the fix, Bob!

PR #116, opened 2021-03-25
Fix ws read loop to handle a pause in the input stream mid-u64 read

Lewis-Kelley:
Add new test to check the fix by mocking such a pause in the stream

**Fixes**

Fixes problem with Swish's websocket implementation where a pause in the stream mid-u64 read would close the connection.

**Proposed changes**

* Switch to using `get-bytevector-n!` instead of `get-bytevector-some!` to wait through a pause in the stream


burgerrg (approved):
Thanks for finding and fixing this bug!

PR #117, opened 2021-03-30
Improve Logging Performance

laqrix:
**Fixes**
In Swish versions up through 2.0.2, a flood of db:log calls can cause
the db gen-server process to fall behind and slow the application. This
deluge can create an out-of-memory condition that crashes the
application.

**Proposed changes**
To remedy this, we add osi_bulk_execute, which takes a vector of SQLite
statements and a vector of bindings, and requests that the database
worker thread do all of the SQLite binding and stepping of statements
before returning. The db gen-server now flushes its queue of db:log
statements using osi_bulk_execute.

We define a "marshaled bindings" C struct and a corresponding Scheme
record type. Each marshaled bindings record represents a complete
set of values that the database worker thread will bind to the
statement.

In practice, the new code took a specific Swish application from an
out-of-memory failure condition, where the database could not keep
up, to maintaining log performance with less than 50 rows per
transaction, where less than 5 is most frequent.

* add osi_bulk_execute
* add "marshaled bindings"
* utilize new functionality in database gen-server for logging

### DIFF
--- a/doc/swish/db.tex
+++ b/doc/swish/db.tex
@@ -196,8 +196,8 @@ processes the following message:
 
 \antipar\begin{itemize}
 
-\item \code{\#(log \var{sql} \var{bindings})}: Add this tuple to
-  the queue. Process the queue.
+\item \code{\#(<log> \var{sql} \var{bindings})}: Add this tuple to the
+  queue. Process the queue.
 
 \end{itemize}
 
@@ -232,6 +232,13 @@ transaction is complete.
 A commit threshold of 10,000 was chosen because it was large enough to
 minimize the cost of a transaction but small enough to execute simple
 queries in less than one second.
+
+Version 2.1.0 adds the concept of marshaled bindings. Bindings are
+copied into the C heap. The resulting handle is wrapped in a Scheme
+record and registered with a guardian via
+\code{make-foreign-handle-guardian}. The database worker process uses
+marshaled bindings to invoke \code{sqlite:bulk-execute} when processing log
+messages.
 
 \section {Programming Interface}
 
@@ -323,8 +330,8 @@ The \code{db:filename} procedure calls \code{(gen-server:call
 \code{ok}
 
 The \code{db:log} procedure calls \code{(gen-server:cast \var{who}
-  \#(log \var{sql} \var{bindings}))}.  \var{sql} is a SQL string, and
-\var{bindings} is a list of values to be bound in the query.
+  \#(<log> \var{sql} \var{bindings}))}.  \var{sql} is a SQL string,
+and \var{bindings} is a list of values to be bound in the query.
 Because \code{db:log} does not wait for a reply from the server, any
 error in processing the request will crash the server.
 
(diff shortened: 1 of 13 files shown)


doc/swish/db.tex
```diff
@@ -593,6 +633,16 @@
...
 
+\defineentry{sqlite:bulk-execute}
+\begin{procedure}
+  \code{(sqlite:bulk-execute \var{stmts} \var{mbindings})}
+\end{procedure}
+\returns{} unspecified
+
+The \code{sqlite:bulk-execute} procedure extracts the handles of the
+vectors \var{stmts} and \var{mbindings} and calls
+\code{osi\_bulk\_execute}.
```
burgerrg: Please add a sentence like _stmts_ is a vector of statement record instances, and _mbindings_ is a vector of corresponding marshaled binding record instances.

doc/swish/db.tex
```diff
@@ -626,11 +676,18 @@
 \end{procedure}
 \returns{} a list of rows where each row is a vector of data in column order
 
-The \code{sqlite:execute} procedure calls \code{(sqlite:bind
-  \var{stmt} \var{bindings})} to bind any variables and then
-iteratively calls \code{(sqlite:step \var{stmt})} to build the
-resulting list of rows. It resets the statement when the procedure
-exits.
+The \code{sqlite:execute} procedure binds any variables in \var{stmt}
```
burgerrg: binds any variables in statement record instance \var{stmt}

doc/swish/db.tex
```diff
@@ -626,11 +676,18 @@
...
+and then calls \code{(sqlite:step \var{stmt})} repeatedly to build the
+resulting list of rows.
+If \var{bindings} is a marshaled bindings record instance,
+then \code{sqlite:execute} calls \code{osi\_bind\_statement\_bindings}
+to bind variables in the statement.
+Otherwise it calls \code{sqlite:marshal-bindings}
+before binding the variables and calls \code{sqlite:unmarshal-bindings}
+before returning.
+As a result, \code{sqlite:execute} accepts as \var{bindings}: a list,
```
burgerrg: accepts the following as \var{bindings}: a list,

burgerrg: Or remove the colon after \var{bindings}

src/swish/db.ss
```diff
@@ -104,7 +111,10 @@
     (gen-server:call who 'filename))
 
   (define (db:log who sql . bindings)
-    (gen-server:cast who `#(log ,sql ,bindings)))
+    (gen-server:cast who
+      (<log> make
+        [sql sql]
+        [mbindings (sqlite:marshal-bindings-no-check bindings)])))
```
burgerrg, line 117: sqlite:marshal-bindings-no-check will create a record for empty bindings, but sqlite:marshal-bindings won't.

Are you sure sqlite:marshal-bindings should return #f when bindings is empty? sqlite:bulk-execute doesn't currently work with #f, so how could one bulk-execute statements without bindings using the sqlite: interface?

src/swish/sqlite.c
```diff
@@ -311,12 +337,167 @@ ptr osi_finalize_statement(uptr statement) {
...
+    return UV_EINVAL;
+  }
+}
+
+ptr get_binding(binding_t* b) {
+  switch (b->type) {
+  case SQLITE_NULL:
+    return Sfalse;
+  case SQLITE_INTEGER:
+    return Sinteger(b->i);
```
burgerrg: use Sinteger64 in case this is compiled on a 32-bit system

doc/swish/db.tex
laqrix: Will add a sentence. Using the phrasing "marshaled **bindings** record instances".  I believe we are using plural bindings throughout for this record type.

doc/swish/db.tex
laqrix: Updating as suggested.

doc/swish/db.tex
laqrix: Removing colon as suggested.

src/swish/sqlite.c
laqrix: Updating as suggested.

src/swish/db.ss
laqrix: Good catch!

To address this, Oscar moved the #f handling into `sqlite:marshal-bindings-no-check` and also added null pointer checks in the C side.  Hopefully that will be robust even if someone gets creative in how they call the API.

burgerrg (approved):
Thanks for this helpful update!

laqrix:
Thanks for the review Bob!

PR #120, opened 2021-05-03
update for glibc 2.33

owaddell:
**Fixes**

* Fix build error "`mallinfo` is deprecated" with glibc 2.33.
* Fix sqlite3.c compiler warning with gcc 11.1.1 on a6le.
* Fix sqlite3.c warnings are not disarmed when `CC='ccache gcc'`.

**Proposed changes**

* Use `mallinfo2` if glibc version >= 2.33.
* Add `-Wmisleading-indentation` to the set of warnings we mute for sqlite3.c.
* Unquote `"${CC}"` in x86* Linux makefiles so we still tame sqlite3.c warnings when `CC='ccache gcc'`.


burgerrg (approved):
Thanks for this update!

owaddell:

PR #121, opened 2021-05-04
mat updates

owaddell:
**Fixes**

* Fix missing index entries for procedures whose names contain `!`.
* Fix `swish-test` appears to generate a broken HTML report when harvesting `.mo` results generated by `run-mat` script.

**Proposed changes**

* Manually escape `!` in index entries. Perhaps we can instead / eventually address this with a TeX macro so we don't have to remember to do this in the future.
* Initialize `.mo` meta-data in `run-mat` so `swish-test` can `--harvest` and generate working HTML `--report`.
* Log `uname` info in the mat meta-data.
* Log the `erlang:now` timestamp for test start/end time in the mat-result meta-data.
* Add a mechanism for logging annotations in the mat-result meta-data recorded in the JSON mat report format.


src/swish/mat.ss
```diff
@@ -241,7 +290,8 @@
           (flush-output-port mo-op))
         (for-each
          (lambda (mat/name)
-           (run-mat mat/name reporter incl-tags excl-tags))
+           (parameterize ([mat-annotations '()])
```
burgerrg: Should the parameterize of `mat-annotations` be done inside `run-mat` so that `run-mat` can be called from the outside world on mats that use annotations?

owaddell: I pushed a commit that moves the `parameterize`. Since we don't yet export `get-result-meta-data`, the `reporter` procedure passed to `run-mat` doesn't have a good way to harvest the annotations or the other result meta-data at the moment.

I didn't want to make a breaking change to `reporter`. An earlier draft used `procedure/arity?` to check for legacy `reporter` interface, but that seemed awkward.

burgerrg: That's OK for now. By moving the parameterize, run-mat will work for mats with annotations, even if there's no good way to get the annotations out.

burgerrg (approved):

burgerrg:
Thanks for these useful improvements!

src/swish/mat.ss
owaddell: Good thought.

owaddell:
Thanks for the review, Bob!

owaddell:
Oops. Looks like I should have force-pushed before merging to avoid confusing Github.

PR #122, opened 2021-05-12
release Swish 2.2.0

owaddell:


burgerrg (approved):
Great work! I'd also support renaming insert-q to @insert-q, remove-q to @remove-q, and changing the call site in $receive to use no-interrrupts around @remove-q.

burgerrg (approved):

owaddell:
@burgerrg, thanks for the quick reviews.

What do you think about the new WIP commits added here? The `ld` call in Mf-arm32le looked broken.

This is related to the Chez Scheme [cross-compile PR](https://github.com/cisco/ChezScheme/pull/571). To cross-compile Swish for arm32le on a6le in an Ubuntu 20.04 VM (continuing from where the instructions in the Chez Scheme PR leave off):
1. `sudo apt install libsystemd-dev:armhf`
2. `git clean` or `make pristine` in the Swish repository so we don't get the wrong sqlite3.o.
3. delete or move the libuv/build dir aside
4. ./configure --scheme=PATH_TO_ARM32_SCHEME --toolprefix=arm-linux-gnueabihf- CFLAGS=-marm
5. make

I get two failures that may be related to my emulation environment: one in `handlers` from http.ms and one in `try-import` in swish-build.ms.

burgerrg:
TOOLPREFIX isn't strictly necessary since we can already configure CC and LD, so it's for convenience in cross-compiling?

owaddell:
Right. It's a convenience inspired by mflatt's 2649600c68ff57efb63d6d5d10c9d9f73368f59a in the Chez Scheme repository.

burgerrg (approved):
The toolprefix looks useful, thanks!

laqrix:
It's still not clear what's happening with Travis. I was able to successfully build and test both normal and coverage configurations using Chez Scheme 9.5.4 and commit `d1d2070` on macOS 11.3.

burgerrg (approved):
Great work; thanks again!

owaddell:
Thanks for the reviews, Bob (and Chris behind the scenes)!

PR #123, opened 2021-05-17
Apply LibUV barrier-err fix

laqrix:
See how Travis handles https://github.com/libuv/libuv/pull/3162


burgerrg:
It's still failing in foreign.ms. Any idea why?

laqrix:
I do not. This was just an attempt to try Oscar's theory.  I don't really understand the nature of the Travis VMs.  My gut feeling is that the filesystem handling is slightly different than real hardware.  In the http cases, we might be missing a file watcher event.  In the foreign.ms case, we might be writing to the filesystem, then proceeding before the write has actually completed.  I think we're stabbing in the dark a bit.

laqrix:
This did not resolve the particular build problem. Closing.

PR #124, opened 2021-07-01
GitHub Actions for tests

burgerrg:
Run builds and tests in GitHub Actions instead of Travis-CI


owaddell (commented):
This is really nice, Bob! Thanks for blazing the trail.

I pushed a couple of commits to consider. One is an alternative to the cast in sqlite.c. The other is a sketch at archiving coverage reports for MacOS, Linux, and Windows. I'm not entirely sold on my use of `main` as a proxy for an explicit "record coverage report for this build" variable we might add to the matrix.

It's not immediately obvious to me why we want to force a subshell in the Mf-a6nt and Mf-i3nt. Is it worth clarifying the comments there?

I wonder if the default 90-day retention policy for archived artifacts is suitable for all builds or whether that is another thing we might want to customize via the matrix configuration.

burgerrg:
Since the lower-level SQLite functions need a 32-bit integer, we’re stuck dealing with it. I'd prefer to cast the vector-length to int in 2 places than do the other approach. We could even raise an error on the Scheme side if the length is too large, but I don't think this is necessary given the limit of 2^31-1.

owaddell:
Thanks for the updates, Bob! Looks good to me.

owaddell (approved):
Thanks again, Bob, for blazing the trail.

PR #125, opened 2021-09-17
test: add "Download Data" link to HTML test report

owaddell:
When generating HTML reports, the swish-test script now inserts a link to download the JSON data. In Firefox, attempting to save the resulting data fails the first time, but succeeds upon retry.

**Fixes**

**Proposed changes**

* Add a "Download Data" link in HTML test reports to provide an easy way to extract the JSON test report data.


burgerrg (approved):
Nice!

owaddell:

PR #126, opened 2021-09-18
add record-type-hash-procedure for pcb and for mon

owaddell:
**Fixes**

O(n) lookup when `equal-hash` is used (with default `record-hash-procedure`) for a functional hash table storing processes or monitors.

**Proposed changes**

Add record-type-hash-procedure for pcb and mon records to support the
use of equal-hash for functional hash tables that hold processes and
monitors. This helps developers avoid poor hashing behavior from the
default record-hash-procedure.

### DIFF
--- a/src/swish/erlang.ms
+++ b/src/swish/erlang.ms
@@ -3753,6 +3753,28 @@
        ,@stack-info)
      (split (get-output-string os) #\newline))))
 
+(isolate-mat record-hash ()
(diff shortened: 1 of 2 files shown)


burgerrg (approved):
Thanks, I wasn't aware of the default behavior of equal-hash on records.

owaddell:
Thanks for the review, Bob!

PR #127, opened 2021-09-21
log os-pid on startup; extend memory statistics

owaddell:
**Fixes**

N/A

**Proposed changes**

* Log the operating-system PID on startup to help correlate Swish logs with other system logs.
* Add current-memory-bytes and maximum-memory-bytes to the statistics we log.

**Question**

* Is it reasonable to call `reset-maximum-memory-bytes!` at each `<statistics>` event?


burgerrg (approved):
Thanks for these useful updates!

burgerrg:
The debug web page update looks good.

laqrix (approved):
This looks good Oscar. Thanks!

owaddell:
Thanks for the reviews!

PR #128, opened 2021-10-05
log additional memory information

owaddell:
**Proposed changes**

* log total physical memory in `<system-attributes>` via `uv_get_total_memory`
* log free memory in `<statistics>` via `uv_get_free_memory`
* add free and total physical memory to the diagnostics debug page

### DIFF
--- a/doc/swish/osi.tex
+++ b/doc/swish/osi.tex
@@ -289,7 +289,7 @@ See Figure~\ref{fig:service-callbacks} for information on these callbacks.
 \caption{Service callbacks. \label{fig:service-callbacks}}
 \end{figure}
 
-\subsection {System Functions}
+\subsection {System Functions and Procedures}
 
 \defineentry{osi\_get\_argv}
 \begin{function}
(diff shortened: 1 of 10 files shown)


doc/swish/osi.tex
```diff
@@ -310,6 +310,25 @@
 function. On macOS, it calls the \code{mstats} function. On Windows,
 it calls the \code{\_heapwalk} function.
 
+\defineentry{osi\_get\_free\_memory}
+\begin{procedure}
+  \code{(osi\_get\_free\_memory)}
+\end{procedure}
+\returns{} an unsigned integer
+
+Returns the number of bytes of free memory as reported by
```
burgerrg: For consistency, "The \code{osi\\_get\\_free\\_memory} function returns ..."

doc/swish/osi.tex
```diff
@@ -310,6 +310,25 @@
...
+\code{uv\_get\_free\_memory}.
+
+\defineentry{osi\_get\_total\_memory}
+\begin{procedure}
+  \code{(osi\_get\_total\_memory)}
+\end{procedure}
+\returns{} an unsigned integer
+
+Returns the number of bytes of total physical memory as reported by
```
burgerrg: See comment on osi_get_free_memory.

doc/swish/statistics.tex
```diff
@@ -108,6 +108,7 @@
   \argrow{gc-real}{elapsed time in seconds of garbage collections
     since last event}
   \argrow{gc-bytes}{Scheme heap bytes reclaimed since last event}
+  \argrow{os-free-memory}{current free memory from \code{osi\_get\_free\_memory} }
```
burgerrg: Suggestion: remove extra space before final }

doc/swish/osi.tex
owaddell: Fixed.

doc/swish/osi.tex
owaddell: Fixed. Thanks!

doc/swish/statistics.tex
owaddell: Fixed.

laqrix:
Question: Would it be wise to wrap the uv_* functions with osi_* wrappers so that if the type signatures of the uv_ functions ever change, the C compiler might notice and throw an error or warning?

burgerrg:
Since the return type is uint64_t, I think it's unlikely for the libuv memory functions to change the return value in the foreseeable future. 2^64 is a gigantic amount of memory.

burgerrg (approved):

owaddell:
I agree the return and argument types for these libuv functions are unlikely to change. I pushed a test that may alert us if they change the units.

I considered adding the new memory information to the default mat meta-data but decided against that since folks can do this with more finesse using `mat:add-annotation!`.

owaddell:
Thanks for the reviews, gang!

PR #129, opened 2021-10-19
add the missing "LD" option to configure file

AiziChen:


burgerrg (approved):
Thank you for this update!

PR #131, opened 2021-11-07
swish-test updates and tilde-expansion for filesystem operations

owaddell:
**Proposed changes**

Add `swish-test` options to make it easier for folks to generate a text-based summary of prior test results or to re-run prior tests that failed or were skipped. Handle limited (current-user) form of tilde expansion for filesystem operations.

* `swish-test`: allow `--progress` in combination with `--harvest` or `--load-results`.
* `swish-test`: add `--rerun` option
* `swish-test`: include exit reason when reporting failure to load profile data
* add `osi_get_home_directory`
* support `~` and paths with `~/` prefix in filesystem procedures

### DIFF
--- a/src/swish/swish-test.ms
+++ b/src/swish/swish-test.ms
@@ -434,7 +434,7 @@
     (format "~a *fail$" (pregexp-quote file3))))
   ;; test --harvest directory processing
   (let ([combined.json (path-combine (output-dir) "combined.json")])
-    (zero-exit
+    (nonzero-exit
      (swish-test "--report " combined.json " --harvest "
        (path-combine (output-dir) test-dir0))
      '("Tests run: 7   Pass: 6   Fail: 1   Skip: 0"))))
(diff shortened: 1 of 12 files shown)


src/swish/swish-test.ms
```diff
@@ -228,8 +240,9 @@
   )
 
 (isolate-mat select-tests ()
+  (define subdir (uuid->string (osi_make_uuid)))
   (define test-file
-    (write-test-file "test-specified.ms"
+    (write-test-file (path-combine "test-specified.ms")
```
burgerrg: Are you missing the path argument to path-combine?

owaddell: Well spotted. I meant to use the subdir to check the case where we `--rerun` all results in a particular directory.

Pushed a fix for that along with some proposed breaking changes.

burgerrg (approved):
Thanks for these useful changes!

laqrix (approved):
Looks good. Thanks Oscar!

owaddell:
Thanks for the reviews, folks!

PR #132, opened 2022-01-17
add Visual Studio 2022

burgerrg:


burgerrg:
We can't use the windows-2022 GitHub Actions test machines because they don't have Visual Studio 2019 for Chez Scheme 9.5.4 and 9.5.6.

burgerrg:
We can use windows-2022 for the Chez Scheme main branch but not for the 9.5.4 or 9.5.6 branches, because they don't support Visual Studio 2022.

laqrix (approved):
This looks good. Thanks Bob!

PR #133, opened 2022-02-01
swish-test: preserve stdout / stderr

owaddell:
**Fixes**

`swish-test` was sending both `stdout` and `stderr` to `stdout`.

**Proposed changes**

Preserve separate output ports in the handler that drains output from spawned OS processes.


burgerrg (approved):

owaddell:
Thanks for the review, Bob!

PR #134, opened 2022-04-12
library updates

burgerrg:
* Update libuv from 1.41.0 to 1.44.1.
  * The CMAKE_RULE_MESSAGES=OFF option is no longer needed.
* Update sqlite from 3.34.1 to 3.38.5.
  * The -DSQLITE_ENABLE_JSON1 option is no longer needed and was removed.
  * The -DSQLITE_OMIT_DESERIALIZE option was added.
  * The log-db migrate mat now checks for upgrade-db failures instead of silently ignoring them.
  * SQLite now stores its column type in all capital letters.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,73 +9,73 @@ jobs:
         config:
           - machine: a6osx
             os: macos-11
-            chez: v9.5.4
+            chez: v9.5.6
           - machine: a6osx
             os: macos-11
-            chez: v9.5.6
+            chez: v9.5.8
           - machine: a6osx
             os: macos-11
             chez: main
           - machine: ta6osx
             os: macos-11
-            chez: v9.5.4
+            chez: v9.5.6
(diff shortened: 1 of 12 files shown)


owaddell:
What do you think about excluding the SQLite serialize/deserialize interface [as proposed here](https://github.com/burgerrg/swish/commit/e65188f156fa769a182f2a82dadb8e2fb084bf59)?

This had been optional but was enabled by default in 3.36.0 and Swish doesn't use it.

burgerrg:
It's fine to omit the serialize/deserialize interface. Thanks for suggesting it!

owaddell-beckman:
FWIW, SQLite 3.38.3 was released recently so I'm waiting for the next shoe to drop.

owaddell (approved):
SQLite seems to have been stable for a while at 3.38.5, so let's roll with it.
Thanks for the updates, Bob!

PR #136, opened 2022-10-07
eliminate jQuery

burgerrg:
There are security problems with jQuery 1.4.4, so it is time to get rid of it.


owaddell-beckman:
Thanks for the update, Bob! I pushed a couple very minor suggestions to a like-named branch.

burgerrg:
Thanks for the suggestions, which I incorporated this morning!

owaddell (approved):
Thanks, Bob!

PR #137, opened 2022-10-15
fix issue flagged by gcc 12.2.1 with -fanalyzer

owaddell-beckman:
**Fixes**

Fixes possible null pointer dereference.

**Proposed changes**

Check return result of `malloc_container`.


burgerrg (approved):
Thanks for finding and fixing this missed case!

owaddell:

PR #138, opened 2023-12-07
Fix bugs and update dependencies

owaddell:
This pull request fixes bugs and updates Swish so that it once again builds with the Chez Scheme main branch. It also updates libuv and SQLite.

1.  Documentation
    - Update README to encourage upgrading to Chez Scheme v9.6.4.
2.  CI Workflow:
    - Trim the test configurations in the CI workflow.
    - Upgrade to Chez Scheme v9.6.4 and resolve duplicate matrix entries.
    - Revise the workflow to support the upcoming Chez Scheme 10.x release and to cache that build.
    - Prune extraneous threaded builds.
    - Skip coverage if tests fail.
    - Test on the latest Ubuntu.
    - Fix deprecated Node.js 12 actions.
3.  Code Updates:
    - Update to development version 2.5.0.
    - Reindent Scheme files using Swish-Lint v1.2.0.
    - Fix infinite regression when exception handler attempts to write to a closed `stderr`.
    - Fix exception when clearing a signal handler that was never installed.
    - Make tests more robust.
    - Update libuv to version 1.47.0 and SQLite to version 3.44.2.
    - Update configure to support static library in default Chez Scheme v10.x configuration.
    - Add a compatibility library to help support Chez Scheme v9.6.x or main.
    - Resort to `#3%hashtable-ref` instead of `#3%symbol-hashtable-ref` in JSON library.
    - Handle differences in argument-count error messages between Chez Scheme v9.6.x and v10.x.
    - Deal with changes to the default heap search path.
    - Fix install permissions in the makefile.
    - Rebuild and relink against the new runtime if Chez Scheme revision changes.
    - Add `Mf-arm64le` for use with Chez Scheme main.
    
We don't have the hardware to test it, but we'd be happy to include a `Mf-arm64osx` if you can put one together.


owaddell:
In test.yml I tried to use `fail-fast: true` with an expression for `continue-on-error` that should mean a failure in a threaded build won't take down the other builds in the matrix while a failure in a non-threaded build will do so.

The workflow documentation leads me to believe this is possible, but some failed builds before I rebased the fix for `claim-exceptions` into this PR show that it didn't work as expected.

burgerrg (approved):
Thanks for this great set of updates!

owaddell:
Thanks for the quick review and for the arm64osx changes!

PR #139, opened 2023-12-07
Documentation updates

owaddell:
* add missing copyright notices
* update the copyright notice in swish.tex
* make the check-doc target more precise and provide a way to mark things as explicitly undocumented
* add missing documentation for `procedure/arity?`
* update documentation that referred to `exit-process`, which is not exported
* fix typo
* use `\etc{}` for meta-language ellipses

### DIFF
--- a/doc/swish/gen-server.tex
+++ b/doc/swish/gen-server.tex
@@ -138,8 +138,9 @@ process and yield \code{\#(error \#(bad-return-value \var{other}))}.
   \#(error \var{error})\alt{}%
   ignore}
 
-\code{gen-server:start} behaves the same as \code{start\&link}
-except that it does not link to the calling process.
+\code{gen-server:start} behaves the same as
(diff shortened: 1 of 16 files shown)


doc/swish/gen-server.tex
```diff
@@ -98,7 +98,7 @@
 
 \code{gen-server:start\&link} spawns the server process, links to
 the calling process, registers the server process as \var{name}, and calls
-\code{(init \var{arg} ...)} within that process. To ensure a
+\code{(init \var{arg} \etc{})} within that process. To ensure a
```
burgerrg: I think `\etc` would suffice.

burgerrg (approved):
Thank you for these useful updates!

doc/swish/gen-server.tex
owaddell: True. I'll update it.

owaddell:

PR #140, opened 2023-12-08
Database updates

owaddell:
* Add `db:start` that does not explicitly `link` with the calling process
* Add end time to the `child` view

### DIFF
--- a/src/swish/db.ms
+++ b/src/swish/db.ms
@@ -416,6 +416,8 @@
     [,@stmts-before (statement-count)]
     [#(EXIT #(bad-arg db:start&link "options"))
      (catch (db:start&link #f ":memory:" 'open "options"))]
+    [#(EXIT #(bad-arg db:start "options"))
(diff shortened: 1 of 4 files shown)


src/swish/db.ms
```diff
@@ -652,32 +654,55 @@
...
+
+      (for-each db:stop (list file tmp mem shared1 shared2 init))
+      (catch (remove-file filename)))))
+  (test db:start&link)
+  (test db:start))
 
-    (db:stop file)
-    (catch (remove-file filename)))))
+(db-mat db-link ()
+  ;; Start a db gen-server and monitor it. Send a nonsene message that
```
burgerrg: `nonsene` => `nonsense`

burgerrg (approved):

src/swish/db.ms
owaddell: Well spotted!

owaddell:
Thanks for the review!

PR #141, opened 2023-12-15
Fix memory leaks

owaddell:
This pull request
* fixes a memory leak in an error path of `sqlite:marshal-bindings`
* fixes a memory leak in `console-event-handler`
* fixes a few places where we need to close ports
* adds notes on running Swish with Address Sanitizer

It also
* adds a missing test for `spawn-os-process-detached`
* adds a `-b` option to `run-mat` to simplify running a specific mat using the mat-prereq artifacts
* reorders some definitions to avoid `letrec*` checks
* quotes paths in `make install` targets
* updates the `.gitattributes` file

### DIFF
--- a/src/swish/http.ss
+++ b/src/swish/http.ss
@@ -176,16 +176,16 @@
          #'(handler conn request header params))]))
 
   (define (http-listener:start&link name sup init-port url-handler options)
-    (define-state-tuple <http-listener> tcp-listener cache-manager pid->op)
+    (define-state-tuple <http-listener> tcp-listener cache-manager pid->ports)
     (define (init)
       (process-trap-exit #t)
       (match-let* ([#(ok ,cache-mgr) (cache-mgr:start&link sup options)])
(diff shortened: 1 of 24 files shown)


src/swish/http.ss
```diff
@@ -211,14 +211,16 @@
                              (conn:start&link cache-mgr ip op options)])
                            (http:handle-input conn url-handler options)))))))])
           (monitor pid)
-          `#(no-reply ,($state copy* [pid->op (ht:set pid->op pid op)])))]
+          `#(no-reply ,($state copy* [pid->op (ht:set pid->op pid (cons ip op))])))]
```
burgerrg: Should the name `pid->op` change to `pid->ports`?

burgerrg (approved):
Thanks for getting the address sanitizer to work!

src/swish/http.ss
owaddell: Good thought!

owaddell:

PR #142, opened 2023-12-18
Websocket improvements

owaddell:
This pull request:
* fixes an issue where we could close a WebSocket due to ping timeout while reading fragmented payloads (To do this, we add a `port->notify-port` mechanism that allows us to monitor the results of `w!` or `r!` for TCP custom ports.);

* provides more accurate `ws:closed` codes when closing a WebSocket;

* adds a test to verify that we process WebSocket pong messages;

* adds a `not-reached` macro that will help us clean up profile coverage output by explicitly calling out cases we should not reach;

* leverages recently added Chez Scheme parameters, if available, to improve both file and WebSocket performance;

* adds infrastructure for using optimized unsafe primitives while adding a CI build that tests with safe versions of these primitives (naturally such code requires careful review, but the ability to test with safe primitives is an improvement over hard-coding `#3%` primitives),

* adds a way to disable Nagle's algorithm and uses this for WebSocket,

* fixes code that assumed `get_bytevector-n` would return as many bytes as requested,

* reduces copying by masking WebSocket payloads in place,

* tries to reduce copying by reusing the first payload as codec buffer,

* adds and uses `osi_tcp_write2` to send the header and payload in a single system call without copying.

### DIFF
--- a/src/swish/websocket.ss
+++ b/src/swish/websocket.ss
@@ -207,6 +207,16 @@
         [127 (get-u64 ip)]
         [,len len]))
 
+    (define (get-bytevector-exactly-n ip size)
+      (let ([bv (make-bytevector size)])
+        (let lp ([i 0])
+          (let ([count (get-bytevector-n! ip bv i (fx- size i))])
+            (if (eof-object? count)
+                (throw 'unexpected-eof)
+                (let ([next (fx+ i count)])
+                  (unless (fx= next size) (lp next))))))
+        bv))
+
     (define (read-frame ip limit)
       (let* ([fin&opcode (get-u8 ip)]
              [final? (fxbit-set? fin&opcode 7)]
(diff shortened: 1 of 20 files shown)


src/swish/websocket.ss
```diff
@@ -246,21 +302,47 @@
...
+      (define default-buf-size 1024)
       (define (if-eof x value)
         (if (eof-object? x)
             value
             x))
-      (let ([ip (payloads->input-port (reverse payloads))])
+      (let ([ip (payloads->input-port payloads)])
         (if (eq? type 'text)
-            (if-eof (get-string-all (binary->utf8 ip)) "")
+            (match-let* ([(,bv0 . ,_) payloads])
```
burgerrg: Is this just `(let ([bv0 (car payloads)]) ...)`?

burgerrg (approved):
Thanks for the significant updates!

src/swish/websocket.ss
owaddell: You're right, that is clearer.

owaddell:
Bob, you're a force of Nature!

PR #143, opened 2023-12-19
export and simplify get-bytevector-exactly-n

owaddell:
Export `get-bytevector-exactly-n` since there were other places where we tried to read a fixed field of _n_ bytes off the wire without checking the length of the result returned by `get-bytevector-n`.

We can simplify the code extracted from websocket.ss since Chez Scheme already handles the looping internally. We just need to check the result.


burgerrg (approved):
Thanks!

owaddell:
Thanks for the quick review!

PR #144, opened 2023-12-19
Adapt define-syntactic-monad

owaddell:
This pull request adapts `define-syntactic-monad` from stex for use in Swish. This is useful, for example, when writing state machines.


burgerrg:
What do you think of putting define-syntactic-monad into meta.ss? I'm OK with it there as well as in its own file.

owaddell:
We considered that and opted to keep dsm.ss in a separate file since the header is different.

burgerrg (approved):
Looks good!

owaddell:
Thanks for the review, Bob!

PR #145, opened 2023-12-20
add natural-string<? and natural-string-ci<?

owaddell:
This pull request adds predicates that compare strings in a way that is more natural to humans.

### DIFF
--- a/doc/swish/erlang.tex
+++ b/doc/swish/erlang.tex
@@ -2980,6 +2980,28 @@ of the elements of list \var{ls} separated by displaying
 the last separator.
 
 % ----------------------------------------------------------------------------
+\defineentry{natural-string<?}
+\defineentry{natural-string-ci<?}
+\begin{procedure}
+  \code{(natural-string<? \var{s$_1$} \var{s$_2$})}\\
+  \code{(natural-string-ci<? \var{s$_1$} \var{s$_2$})}\strut
+\end{procedure}
+\returns{} a boolean
+
+These procedures compare strings \var{s$_1$} and \var{s$_2$} in a way that is
+more natural to humans than the corresponding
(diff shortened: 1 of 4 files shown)


doc/swish/erlang.tex
```diff
@@ -2979,6 +2979,29 @@ \subsection{String Utilities}
 \var{separator}. When \var{last-separator} is specified, it is used as
 the last separator.
 
+\pagebreak
```
burgerrg: This manual page break may introduce a maintenance burden.

doc/swish/erlang.tex
```diff
@@ -2979,6 +2979,29 @@ \subsection{String Utilities}
...
+
+These procedures compare strings \var{s$_1$} and \var{s$_2$} in a way that is
+more natural to humans than the corresponding
+\code{string<?} and \code{string-ci<?} predicates.
+The natural comparison predicates:
+\begin{enumerate}
+ \item ignore leading and trailing whitespace
+ \item treat runs of multiple whitespace characters as a single space
+ \item compare integers numerically, treating \code{-} as a dash
+   if it follows an alphanumeric character or else as negation
```
burgerrg: consider using a comma at the end of items 1 & 2, and ", and" at the end of item 3.

burgerrg (approved):
This looks handy!

doc/swish/erlang.tex
owaddell: Good catch. I left that in by mistake.

doc/swish/erlang.tex
owaddell: Much better!

owaddell:
Thanks for the speedy review!

PR #146, opened 2023-12-20
Profiling updates

owaddell:
**Fixes**

* Fix the destination for annotated files when generating a coverage report where some source files are included via relative path.

**Proposed changes**

* Improve coverage report's rendering of annotated source code:
   - Add line numbers.
   - Do not color leading whitespace.
   - Change DOCTYPE to HTML5.
* Add a coverage legend within annotated source files.
* Tidy the HTML output in profile report (and in `swish-test`'s HTML mat output)
* Adapt to a change from version 9.5.1 of Chez Scheme that maintains profile counts even when the collector reclaims source code.
* Strip the common prefix from source files in the coverage summary and sort based on real path.


src/swish/profile.ss
```diff
@@ -464,36 +604,113 @@
...
+        "<meta http-equiv=\"Content-Type\" content=\"text/html; charset=UTF-8\">"
+        "<title>~a</title>"
+        "<style type=\"text/css\">"
+        "<!--\n") name)
+    (fprintf op "BODY { font-family: monospace; }\n")
+    (fprintf op "span:hover { outline: 1px solid darkgray; }\n")
+    (fprintf op "table.code { padding:0; border-spacing:0; white-space: pre; }\n")
+    (fprintf op "table.code td { padding-top:0; padding-bottom:0; }\n")
+    ;; For convenince when copy/pasting from the rendered output, the
+    ;; line numbers should not be selectable.
```
burgerrg, line 622: Nice idea!

burgerrg (approved):
I like the new style!

src/swish/profile.ss
burgerrg: `convenince` => `convenience`

burgerrg:
According to the HTML5 validator:

- The type attribute for the style element is not needed and should be omitted.
- The summary attribute on the table element is obsolete. Consider describing the structure of the table in a caption element or in a figure element containing the table; or, simplify the structure of the table so that no description is needed.

src/swish/profile.ss
owaddell: I agree. Well done, Chris!

owaddell:
We made some updates to address HTML5 validator warnings. Turns out they have a container image that makes it a bit handier. There is a CSS warning that looks like a known bug in the validator.

We also ran HTML Tidy 5.8.0.

burgerrg (approved):
Nice work!

owaddell:
Thanks for the reviews, Bob!

PR #148, opened 2024-01-04
add osi_is_service

owaddell:
Add `osi_is_service` so applications can check whether or not they are running as a service.


burgerrg (commented):
I don't see where g_is_service is set to 0 when it's not a service.

owaddell:
We're relying on C standard initialization for objects with static storage duration.

burgerrg (approved):

owaddell:
Thanks for the review, Bob!

PR #149, opened 2024-01-05
Add file utilities

owaddell:
Add `filter-files` and `fold-files` since we often reimplement this functionality in other projects.


burgerrg (approved):
Thanks for these useful functions!

owaddell:
Thanks for the review, Bob!

PR #150, opened 2024-01-09
update swish-build and swish-test

owaddell:
`swish-build` updates:
- fix `-s` to work for the _source_ file
- add `--prelude` and `--prelude-file` for more control over compilation
- clarify `--help` text

`swish-test` updates:
- use case-sensitive compare for file extensions
- use new `filter-files`
- add `--progress suite-verbose` option and use it in our build scripts

### DIFF
--- a/src/run-mats
+++ b/src/run-mats
@@ -15,7 +15,7 @@ fail() {
 }
 
 case $# in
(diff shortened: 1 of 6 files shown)


burgerrg (approved):
Thanks for these updates!

owaddell:
Thanks for the review, Bob!

PR #153, opened 2024-01-18
JSON enhancements

owaddell:
This pull request
* improves JSON read/write performance
* adds a `json:key<?` process parameter to control sorting of fields when writing JSON objects; if `#f` the output order is arbitrary
* adds `json:pretty` which is more convenient than `json:write` for common cases
* adds a `json:custom-write` process parameter to supply the optional `custom-write` argument for JSON output procedures


doc/swish/http.tex
```diff
@@ -1134,6 +1134,20 @@ \subsection{WebSocket Protocol}\label{sec:websocket-protocol}
...
+\defineentry{json:key<?}
+\begin{parameter}
+  \code{json:key<?}
+\end{parameter}
+\returns{} a boolean or a predicate for comparing two strings
+
+The \code{json:key<?} parameter determines whether and how
+\code{json:write} and \code{json:write-object} sort keys
+when writing JSON objects.
+If set to \code{\#f}, keys are not sorted and the output order is arbitrary.
```
burgerrg, line 1183: Add a comma after "sorted".

doc/swish/http.tex
```diff
@@ -1134,6 +1137,49 @@ \subsection{WebSocket Protocol}\label{sec:websocket-protocol}
...
+\returns{} unspecified
+
+The \code{json:pretty} procedure formats an object \var{x} to the
+optional textual output port \var{op} in a human-readable form. The
+object is indented and keys are sorted using
+\code{natural-string-ci<?}. The default value of \var{op} is
+\code{(current-output-port)}.
+
+Calling \code{(json:pretty x)} is equivalent to calling
+\code{(json:write (current-output-port) x 0)}.
```
burgerrg: It's equivalent except that it uses natural-string-ci<? instead of string<?.

doc/swish/http.tex
```diff
@@ -1106,6 +1106,9 @@ \subsection{WebSocket Protocol}\label{sec:websocket-protocol}
 objects.  The \var{wr} procedure is equivalent to
 \code{(lambda (\var{op} \var{x} \var{indent}) (json:write \var{op} \var{x} \var{indent} \var{custom-write}))}.
 
+The default value of \var{custom-write} is the value stored in the
```
burgerrg, line 1112: The description of sorting the keys by string<? needs to be updated to talk about the new process parameter.

Check the description of json:write-object, too.

src/swish/json.ss
```diff
@@ -620,15 +620,26 @@
         (let-values ([(op get) (open-string-output-port)])
           (write-string (json-key->string k) op)
           (get)))
+      (define (format-value wr val)
+        (and (free-identifier=? wr #'json:write)
+             (or (boolean? val) (fixnum? val) (string? val))
```
burgerrg: Do you want to include `(and (flonum? val) (finite? val))`?

burgerrg: Do you want to include `(eqv? val #\nul)`?

doc/swish/http.tex
owaddell: Fixed, thanks.

doc/swish/http.tex
owaddell: Good point. Revised to clarify.

doc/swish/http.tex
owaddell: Good find. I tried to update the relevant places.

src/swish/json.ss
owaddell: We may as well. I pushed an update to do this (stopping short of bignums) and added some tests.

burgerrg (approved):

owaddell:
Thanks for the timely reviews, Bob!

PR #154, opened 2024-01-23
Another attempt at more robust macOS test builds

owaddell:
This pull request
* updates the database `commit-delay` test to log debugging information if the test fails,
* scales more timeouts in the HTTP mats,
* and uses `test:receive` to report the time elapsed in `receive` before hitting a clause or timeout.

If these changes look okay, we may want to consider pulling this in ahead of the JSON changes where macOS test failures seem to have caused problems for some of the builds.

### DIFF
--- a/src/swish/db.ms
+++ b/src/swish/db.ms
@@ -1065,10 +1065,26 @@
     [,_ (event-mgr:notify `#(,self waiting))]
     [ok (receive [in-sync 'ok])]
     [,@expected (reverse rev-actual)])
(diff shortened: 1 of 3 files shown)


burgerrg (approved):
The changes look good. It'll be fun to see what we learn when GitHub actions take too long to run.

owaddell:
Thanks for the quick review. I'm curious as well to see if we learn anything. Hopefully the scaled timeouts make failures less frequent.

PR #156, opened 2024-02-04
clean up test-request shared library

owaddell:
Remove the test-request shared library on `make clean`. Failing to do this meant that the osi.ms submit-request mat could fail when switching architectures despite a `make clean` before switching.


burgerrg (approved):
Thanks for figuring this out!

owaddell:
Thanks, Bob!

PR #157, opened 2024-02-07
Test fixes

payneca:
**Fixes**

* Fixes an issue with running a compiled app under Chez v10. The test would fail if Chez is installed using `make install`. The build server sets the `PATH` and does not expose the problem.
* Updates isolate-mat's `cleanup-after` to use shutdown escalation similar to supervisors. Upcoming changes revealed that there can be rogue process that still don't die after killing with a `shutdown` signal.


burgerrg (approved):
Thanks for these changes!

payneca:
Thanks for the _very_ quick review Bob!

PR #158, opened 2024-02-15
add parallel interface

payneca:
**Proposed changes**

The parallel interface provides an API for spawning multiple worker processes and collecting results in a fault-tolerant system. Because Swish is single-threaded using lightweight processes, not all programs benefit from multiplexing using this mechanism. Programs that are I/O bound or spawn external OS processes benefit the most.

This commit adds the following exports: `parallel`, `parallel!`, `parallel:execute`, `parallel:execute!`, `parallel:for-effect`, `parallel:map`, and `parallel:vector-map`


doc/swish/parallel.tex
```diff
@@ -0,0 +1,183 @@
+% Copyright 2023 Beckman Coulter, Inc.
```
burgerrg, line 1: 2024? Maybe now is a good time to change the 2023 to 2024 in doc/swish.tex?

doc/swish/parallel.tex
```diff
@@ -0,0 +1,183 @@
...
+are shown in Figure~\ref{fig:parallel}.
+
+\begin{figure}
+  \center\includegraphics{swish/parallel-theory.pdf}
+  \caption{\label{fig:parallel}Parallel kernel}
+\end{figure}
+
+% Note: This description is very similar to a comment in the code. Try
+% to keep them in sync.
+The desired behavior here is a uni-directional link which is not
```
burgerrg: add a comma after "link"

doc/swish/parallel.tex
```diff
@@ -0,0 +1,183 @@
...
+native to Swish. To accomplish this, the caller starts the kernel
+using \code{spawn\&link} and the kernel enables
```
burgerrg: add a comma after "spawn&link"

doc/swish/parallel.tex
```diff
@@ -0,0 +1,183 @@
...
+\code{process-trap-exit}. An exit signal generated by a worker
+propagates to other workers, but does not signal the caller. An exit
+signal from the caller will propagate to the workers and the kernel
```
burgerrg: add a comma after "workers"

doc/swish/parallel.tex
```diff
@@ -0,0 +1,183 @@
...
+will complete normally.
+
+The kernel is configured using \code{parallel:options}. The kernel
+starts up to \code{start-limit} worker processes using
+\code{spawn\&link} in some \code{order}\footnote{Because workers are
+independent processes, start order does not guarantee execution
+order. Control of start order provides a mechanism to help detect and
+test for assumptions and errors in user-code.}. As each worker
```
burgerrg: "user-code" => "user code"

doc/swish/parallel.tex
```diff
@@ -0,0 +1,183 @@
...
+completes, a new one starts. When all the workers complete, the kernel
+returns an API-specific value.
+
+The kernel also supports a \code{timeout} for the entire parallel
+operation. When the timeout expires all remaining workers are killed
```
burgerrg: add a comma after "expires"

doc/swish/parallel.tex
```diff
@@ -0,0 +1,183 @@
...
+with the reason \code{timeout} and an exception is thrown in the
```
burgerrg: add a comma after "timeout"

doc/swish/parallel.tex
```diff
@@ -0,0 +1,183 @@
...
+  \code{(parallel ([\var{option} \var{value}] \etc) \var{e} \etc)}\\
+  \code{(parallel ,\var{options} \var{e} \etc)}
+\end{syntax}
+
+The \code{parallel} construct invokes each expression \var{e}, \etc{}
+concurrently and returns a list of the resulting values. Expressions
+may run concurrently unless \code{start-limit} is 1.
+
+The \code{(parallel ([\var{option} \var{value}] \etc) \var{e} \etc)}
+form constructs a copy of the default options overriding the specified
```
burgerrg: add a comma after "options"
change "the" to "each": overriding each specified option ...

doc/swish/parallel.tex
```diff
@@ -0,0 +1,183 @@
...
+\var{option} with the specified \var{value}.
+
```
burgerrg, line 125: What does the ",options" form do? Does it allow you to specify an expression for the options object to use?

doc/swish/parallel.tex
```diff
@@ -0,0 +1,183 @@
...
+\begin{procedure}
+  \code{(parallel:execute \opt{\var{options}} \var{thunks})}
+\end{procedure}
+\returns{} a list of results
+
+The \code{parallel:execute} procedure takes a list of \var{thunks}
+(procedures of no arguments). \code{parallel:execute} returns a list
+of resulting values. Each \var{procedure} may run concurrently unless
+\code{start-limit} is 1.
+
```
burgerrg, line 150: Describe the optional "options" argument.

doc/swish/parallel.tex
```diff
@@ -0,0 +1,183 @@
...
+\begin{procedure}
+  \code{(parallel:execute! \opt{\var{options}} \var{thunks})}
+\end{procedure}
+\returns{} unspecified
+
+The \code{parallel:execute!} procedure takes a list of \var{thunks}
+(procedures of no arguments). \code{parallel:execute!} returns an
+unspecified value. Each \var{procedure} may run concurrently unless
+\code{start-limit} is 1.
+
```
burgerrg, line 159: Describe the optional "options" argument.

doc/swish/parallel.tex
```diff
@@ -0,0 +1,183 @@
...
+\begin{procedure}
+  \code{(parallel:for-effect \opt{\var{options}} \var{procedure} \var{list$_1$} \var{list$_2$} \etc)}
+\end{procedure}
+\returns{} unspecified
+
+The \code{parallel:for-effect} procedure works like the
+\code{parallel:map} procedure except that it does not accumulate and
+return a list of values. Each application of \var{procedure} may run
+concurrently unless \code{start-limit} is 1.
+
```
burgerrg, line 170: Describe the optional "options" argument.

What do you think of calling it parallel:for-each instead?

doc/swish/parallel.tex
```diff
@@ -0,0 +1,183 @@
...
+\defineentry{parallel:map}
+\begin{procedure}
+  \code{(parallel:map \opt{\var{options}} \var{procedure} \var{list$_1$} \var{list$_2$} \etc)}
+\end{procedure}
+\returns{} a list of results
+
+The \code{parallel:map} procedure is similar to Chez Scheme's
+\code{map} procedure. Each application of \var{procedure} may run
+concurrently unless \code{start-limit} is 1.
+
```
burgerrg, line 186: Describe the optional "options" argument.

doc/swish/parallel.tex
```diff
@@ -0,0 +1,183 @@
...
+return a list of values. Each application of \var{procedure} may run
+concurrently unless \code{start-limit} is 1.
+
+\defineentry{parallel:map}
+\begin{procedure}
+  \code{(parallel:map \opt{\var{options}} \var{procedure} \var{list$_1$} \var{list$_2$} \etc)}
+\end{procedure}
+\returns{} a list of results
+
+The \code{parallel:map} procedure is similar to Chez Scheme's
```
burgerrg: "is similar to" => "works like" for consistency with the other procedures

doc/swish/parallel.tex
```diff
@@ -0,0 +1,183 @@
...
+
+\defineentry{parallel:vector-map}
+\begin{procedure}
+  \code{(parallel:vector-map \opt{\var{options}} \var{procedure} \var{vector$_1$} \var{vector$_2$} \etc)}
+\end{procedure}
+\returns{} a vector of results
+
+The \code{parallel:vector-map} procedure works like Chez Scheme's
+\code{vector-map} procedure. Each application of \var{procedure} may
+run concurrently unless \code{start-limit} is 1.
```
burgerrg, line 195: Describe the optional "options" argument.

src/swish/parallel.ss
```diff
@@ -0,0 +1,378 @@
+;;; Copyright 2023 Beckman Coulter, Inc.
```
burgerrg, line 1: 2024?

src/swish/parallel.ss
```diff
@@ -0,0 +1,378 @@
...
+        [random
+         (do ([i 0 (fx+ i 1)]) ((fx= i len) vec)
+           (vector-set! vec i i))
+         (shuffle-vector! vec)]
+        [left
+         (do ([i 0 (fx+ i 1)]) ((fx= i len) vec)
+           (vector-set! vec i i))]
+        [right
+         (do ([i 0 (fx+ i 1)]) ((fx= i len) vec)
+           (vector-set! vec i (fx- len i 1)))])))
```
burgerrg: Consider an fxvector to avoid the dirty writes.

src/swish/parallel.ss
```diff
@@ -0,0 +1,378 @@
...
+     [(opt thunks)
+      ($parallel-map1 'parallel:execute (check-opt 'parallel:execute opt) thaw thunks)]))
+
+  (define ($parallel-vec1 who opt f v)
+    (scatter/gather! who opt f (vector-copy v) values))
+
+  (define ($parallel-vec* who opt f more)
+    (scatter/gather! who opt
+      (lambda (v*) (apply f (vector->list v*)))
+      (apply vector-map vector more)
```
burgerrg: avoid the conversion of vector to list:
`(lambda (v*) (apply f v*))`
`(apply vector-map list more)`

src/swish/parallel.ss
```diff
@@ -0,0 +1,378 @@
...
+    (check who first)
+    (let ([n (get-length first)])
+      (let lp ([rest rest])
+        (cond
+         [(null? rest) #t]
+         [(begin
+            (check who (car rest))
+            #f)]
+         [(fx= (get-length (car rest)) n) (lp (cdr rest))]
+         [else (errorf who error-msg first (car rest))]))))
```
burgerrg: This was hard to read. Consider the simpler:
```
      (let lp ([rest rest])
        (unless (null? rest)
          (let ([next (car rest)])
            (check who next)
            (unless (fx= (get-length next) n)
              (errorf who error-msg first next))
            (lp (cdr rest)))))
```

src/swish/parallel.ss
```diff
@@ -0,0 +1,378 @@
...
+         [else
+          (check* who f/arg1 args)
+          (proc* who default-options f (cons f/arg1 args))]))]
+     [(maybe-opt opt/f) =>
+      (lambda (opt)
+        (cond
+         [(maybe-proc f/arg1) =>
+          (lambda (f)
+            (cond
+             [(not (pair? args)) (errorf #f "incorrect number of arguments to ~s" who)]
```
burgerrg: (not (pair? args)) => (null? args)

src/swish/parallel.ms
```diff
@@ -0,0 +1,581 @@
+;;; Copyright 2023 Beckman Coulter, Inc.
```
burgerrg, line 1: 2024?

src/swish/parallel.ss
```diff
@@ -0,0 +1,378 @@
...
+    (define (start-work idx)
+      (let* ([i (vector-ref index-vec idx)]
+             [v (vector-ref vec i)]
+             [pid (spawn&link (lambda () (vector-set! vec i (f v))))]
+             [m (monitor pid)])
+        (eq-hashtable-set! monitors m pid)))
+    (define (sg started)
+      (let ([running (hashtable-size monitors)])
+        (cond
+         [(and (fx= started len) (= running 0))
```
burgerrg: You could use fx= here.

src/swish/parallel.ss
```diff
@@ -0,0 +1,378 @@
...
+          vec]
+         [(and (fx< started len) (< running start-limit))
```
burgerrg: You could use fx< here.

src/swish/parallel.ss
```diff
@@ -0,0 +1,378 @@
...
+    (define (wait-for-any expire-time raise?)
+      (receive (until expire-time (throw 'timeout))
+        [`(DOWN ,m ,_ ,reason ,err)
+         (cond
+          [(eq-hashtable-ref monitors m #f)
+           (eq-hashtable-delete! monitors m)
+           (when (and raise? (not (eq? reason 'normal)))
+             (raise err))]
+          [else
+           (wait-for-any expire-time raise?)])]
```
burgerrg: Do you need the monitors since you'll get the EXIT messages?

burgerrg:
```
  (define ($scatter/gather! opt f vec)
    (match-define `(<parallel:options> ,start-limit ,order ,timeout) opt)
    (define me self)
    (define len (vector-length vec))
    (define index-vec (build-index-vector len order))
    (define pid-ht (make-eq-hashtable))
    (define expire-time
      (if (eq? timeout 'infinity)
          'infinity
          (+ (erlang:now) timeout)))
    (define parent-pid (process-parent))
    (define (wait-for-any expire-time raise?)
      (receive (until expire-time (throw 'timeout))
        [`(EXIT ,pid ,reason ,err)
         (cond
          [(and raise? (eq? pid parent-pid))
           ;; When the kernel is running and the parent dies, escape
           ;; so the kernel can stop the workers and die for the
           ;; same reason.
           (raise err)]
          [(eq-hashtable-ref pid-ht pid #f)
           (eq-hashtable-delete! pid-ht pid)
           (when (and raise? (not (eq? reason 'normal)))
             (raise err))]
          [else
           (wait-for-any expire-time raise?)])]))
    (define (start-work idx)
      (let* ([i (vector-ref index-vec idx)]
             [v (vector-ref vec i)]
             [pid (spawn&link (lambda () (vector-set! vec i (f v))))])
        (eq-hashtable-set! pid-ht pid 0)))
    (define (sg started)
      (let ([running (hashtable-size pid-ht)])
        (cond
         [(and (fx= started len) (= running 0))
          vec]
         [(and (fx< started len) (< running start-limit))
          (start-work started)
          (sg (fx+ started 1))]
         [else
          (wait-for-any expire-time #t)
          (sg started)])))
    (match (try (sg 0))
      [`(catch ,_ ,err)
       (let ([pids (hashtable-keys pid-ht)])
         (when (> (vector-length pids) 0)
           (vector-for-each (lambda (pid) (kill pid err)) pids)
           ;; Wait for all workers to complete in case they are
           ;; holding critical resources needed by any subsequent
           ;; user-code.
           (vector-for-each (lambda (pid) (wait-for-any 'infinity #f)) pids)))
       (raise err)]
      [,result result]))
```

doc/swish/parallel.tex
payneca: I explicitly opted for 2023 because the code was created and published to a public fork in 2023.  I will add a separate commit to update doc/swish.tex.

doc/swish/parallel.tex
payneca: Updating as suggested.

doc/swish/parallel.tex
payneca: Updating as suggested.

doc/swish/parallel.tex
payneca: Updating as suggested.

doc/swish/parallel.tex
payneca: Updating as suggested.

doc/swish/parallel.tex
payneca: Updating as suggested.

doc/swish/parallel.tex
payneca: Updating as suggested.

doc/swish/parallel.tex
payneca: Updating as suggested.

doc/swish/parallel.tex
payneca: That is correct. I will add a sentence to clarify.

doc/swish/parallel.tex
payneca: I prefer the "for-each" naming. I like the mechanical "for-each" => "parallel:for-each", but I don't want someone to accidentally believe that parallel:for-each will start or run process in left-to-right order. If you think that won't be an issue, I can rename it.

doc/swish/parallel.tex
payneca: Updated as suggested.

doc/swish/parallel.tex
burgerrg: I don't think that will be a problem. The prefix "parallel" makes it clear that things can happen in any order.

doc/swish/parallel.tex
payneca: Updating as suggested.

doc/swish/parallel.tex
payneca: Updating as suggested.

doc/swish/parallel.tex
payneca: Updating as suggested.

doc/swish/parallel.tex
payneca: Updating as suggested.

doc/swish/parallel.tex
payneca: Updating to describe "options" argument and renaming to "parallel:for-each".

src/swish/parallel.ss
payneca: I used monitors to protect against intentional or accidental use of `unlink` by the subprocess to disconnect from the parallel kernel. If we don't mind the caveat, using links only is simpler.

src/swish/parallel.ss
payneca: Updating as suggested.

src/swish/parallel.ss
payneca: Updating as suggested.

src/swish/parallel.ss
payneca: Updating as suggested.

src/swish/parallel.ss
payneca: Updating as suggested.

src/swish/parallel.ss
payneca: Updating as suggested.

src/swish/parallel.ss
payneca: Updating as suggested.

src/swish/parallel.ss
burgerrg: Thanks for reminding me of why supervisors use monitors in addition to links. I very rarely use unlink, so it seems like a reasonable caveat. I wonder what the memory & time savings would be for eliminating the monitors? I suspect it'll be very small.

burgerrg: I'd say keep the monitors so that the supervision is consistent with the supervisor module.

burgerrg: I didn't read the supervisor code carefully enough. It doesn't use monitors until shutdown time.

burgerrg:
You could use declare-unsafe-primitives to give you a small performance boost in the shuffler and other parts that deal with known fixnums.

src/swish/parallel.ss
payneca: We'll go with the "no monitor" solution. The reduction in code complexity won. There is some mild performance improvement. We currently don't have a lot of "real world" examples to measure.

The very simple, but unlikely `(parallel () (unlink (process-parent)))` will cause `parallel` to never return. Since supervisors and event-mgr are also susceptible to naughty children unlinking, we may want to address that in a future release. It would be nice to have a design pattern we can rely on without storing tables of monitors.

src/swish/parallel.ss
```diff
@@ -0,0 +1,378 @@
...
+       (lambda (x)
+         (or (eq? x 'infinity)
+             (and (fixnum? x) (fxnonnegative? x))))]]))
+
+  (define default-options (parallel:options))
+
+  (include "swish/unsafe.ss")
+
+  (define (shuffle-fxvector! v)
+    (declare-unsafe-primitives fx> fx- fxvector-ref fxvector-set!)
```
burgerrg: fxvector-length can be added, since all callers of shuffle-fxvector! pass an fxvector.

burgerrg (approved):
Nice work!

src/swish/parallel.ss
```diff
@@ -0,0 +1,378 @@
...
+         (do ([i 0 (fx+ i 1)]) ((fx= i len) vec)
+           (fxvector-set! vec i i))]
+        [right
+         (do ([i 0 (fx+ i 1)]) ((fx= i len) vec)
+           (fxvector-set! vec i (fx- len i 1)))])))
+
+  (define ($scatter/gather! opt f vec)
+    (match-define `(<parallel:options> ,start-limit ,order ,timeout) opt)
+    (declare-unsafe-primitives fx+ fx= fx< fxvector-ref
+      vector-ref vector-set!)
```
burgerrg: Consider adding vector-length, eq-hashtable-ref, eq-hashtable-delete!, eq-hashtable-set!, hashtable-size, hashtable-keys

src/swish/parallel.ss
payneca: I had intentionally left it out, but upon review, I think it'll be fine. Updating as suggested.

src/swish/parallel.ss
payneca: I had intentionally left out vector-length, but it will be fine. We thought the compiler would do better with the hashtable primitives, but on review found it didn't quite do what we thought. I'll add them to the unsafe list.

payneca:
Thanks for the detailed review @burgerrg !

PR #159, opened 2024-02-21
add arm64osx automated tests

burgerrg:


.github/workflows/test.yml
```diff
@@ -15,7 +15,7 @@ jobs:
         id: build-clone-main
         run: |
           git clone --filter=blob:none --branch main https://github.com/cisco/ChezScheme.git
-          echo "HASH_KEY=$(cat ChezScheme/.git/refs/heads/main)" >> "${GITHUB_OUTPUT}"
+          echo "HASH_KEY=2$(cat ChezScheme/.git/refs/heads/main)" >> "${GITHUB_OUTPUT}"
```
payneca: Is this `2` a typo, or does it do something I don't know about. Trying it locally seems to just prepend `2` to the hash which doesn't seem correct.

payneca (commented):
In an upcoming commit, we noticed that we aren't using coreutils' `ln` either. Our solution there was to explicitly update `PATH` to catch all of coreutils. Would that be a better solution to the `realpath` problem?

.github/workflows/test.yml
burgerrg: It's important, because without it, the current cache won't have arm64osx in it.

burgerrg:
Yes, updating PATH to include all of coreutils would be fine. Be sure to use $(brew --prefix) to get the proper prefix for both Intel and ARM Macs.

burgerrg:
If you do add "$(brew --prefix)/opt/coreutils/libexec/gnubin" to the PATH, then you should update the README and configure scripts that use ginstall and grealpath.

burgerrg:
@payneca, I removed the changes to README and configure and simply updated GITHUB_PATH in the automated tests. Feel free to update this in a future pull request when you update everything else.

.github/workflows/test.yml
owaddell: It seems the `2` is essentially marking a "schema change" for the cache. What do you think about changing the `2` to something like `v2_` to help us recognize this in the future?

(When I saw a `2` before process substitution, I initially wondered if you were doing something novel with stderr.)

burgerrg: Great idea, done!

burgerrg:
@owaddell, I also changed fail-fast to false so that each machine type gets the opportunity to complete should any of the others fail.

payneca (approved):

owaddell (approved):

PR #161, opened 2024-02-23
update console-event-handler to elide specific <child-end> messages

owaddell:
`<child-end>` messages where the exit reason is `normal` or `shutdown` and have no stack details are expected and are generally not useful. The exiting process may be killed by their supervisor so the killed field is ignored.


burgerrg (approved):
Thanks for the useful simplifying change!

owaddell:
Thanks for the review!

PR #162, opened 2024-02-26
Fixes

owaddell:
- fix typo in mat error messages
- make includes of "unsafe.ss" consistent
- fix a potential crash in `watch-path` callback
- make Swish scripts less broken under Cygwin

### DIFF
--- a/src/swish/app.ss
+++ b/src/swish/app.ss
@@ -132,6 +132,36 @@
                      (software-version 'swish))])])
       (printf "~a~@[ Version ~a~]~@[ (~a)~]\n" name version revision)))
 
+  (meta-cond
+   [windows?
+    (define (split-path p n)
+      (let ([rest (path-rest p)])
+        (if (or (= n 0) (equal? p rest))
+            (list p)
+            (cons (path-first p) (split-path (path-rest p) (- n 1))))))
+    (define (rewrite-path script)
+      ;; Cygwin paths start with /.
+      ;; By default, MSYS2 and Git Bash automatically convert to Windows paths.
+      (match (split-path script 3)
+        [("/" "cygdrive" ,drive ,path)
+         (path-combine (string-append drive ":") path)]
+        [("/" "tmp" . ,more)
+         ;; assume Cygwin's fstab sets a usertemp mount for /tmp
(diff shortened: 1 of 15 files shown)


src/swish/app.ss
```diff
@@ -132,6 +132,35 @@
                      (software-version 'swish))])])
       (printf "~a~@[ Version ~a~]~@[ (~a)~]\n" name version revision)))
 
+  (meta-cond
+   [(memq (machine-type) '(a6nt ta6nt i3nt ti3nt))
```
burgerrg: `(eqv? (directory-separator) #\\)` will include the ARM versions of Windows in the future should we add them.

src/swish/app.ss
```diff
@@ -132,6 +132,35 @@
...
+    (define (rewrite-path script)
+      ;; Cygwin paths start with /.
+      ;; By default, MSYS2 and Git Bash automatically convert to Windows paths.
+      (match (split-path script 3)
+        [("/" "cygdrive" ,drive ,path)
+         (path-combine (string-append drive ":") path)]
+        [("/" "tmp" . ,more)
+         (apply path-combine (osi_get_temp_directory) more)]
+        [("/" . ,_)
+         (let ([root (get-registry "hkey_local_machine\\Software\\Cygwin\\setup\\rootdir")])
```
burgerrg: HKEY_LOCAL_MACHINE is typically used in uppercase.

burgerrg (approved):
Nice updates! The resolve path was a clever way to handle typical Cygwin paths.

src/swish/app.ss
owaddell: That's a cute idea. I'm going to try updating the definition in meta.ss and see if we can replace most of the `(machine-type)` calls. Hopefully the build server will let me know if this was misguided.

src/swish/app.ss
owaddell: Okay, I'll try it in uppercase. The CSUG `get-registry` entry may have misled me.

burgerrg (approved):
Adding `windows?` worked really well!

owaddell:
Thanks for the review and improvements!

PR #163, opened 2024-02-28
Handling trapped EXIT and interrupts

owaddell:
Fix an issue with `make install` permissions on Windows and make it work for the `mat-prereq` target as well.

Support a procedural handler as the value for `process-trap-exit` so that a process blocked in a gen-server call can perform some clean-up if killed.
    
This commit also fixes an obscure bug in the case where a `receive` form has a `guard` that calls `receive`.
    
Rework the interactive console so that:
- we can interrupt the REPL via ^C on Windows again,
- we can wait until interrupts are enabled before invoking the keyboard-interrupt-handler, and
- a script can start an interactive REPL that handles ^C as Swish does.
    
User visible changes to the behavior of the interactive console include:
- trying to read from the console in a spawned process returns `#!eof`,
- ~calling `(receive [,x x])` at the REPL can intercept a `#<cmsg>` if the user enters a line of input while the receive is running; if the user does not try to type ahead, the `receive` works as expected.~ (reworked console to address this)


doc/swish/erlang.tex
```diff
@@ -973,11 +973,41 @@
...
+\code{gen-server:call}.
+
+% ----------------------------------------------------------------------------
+\defineentry{raise-on-exit}
+\begin{procedure}
+  \code{(raise-on-exit \var{msg})}
+\end{procedure}
+\returns{} see below
+
+The \code{raise-on-exit} procedure expects an exit message
```
burgerrg: expects an exit message \var{msg} ...

doc/swish/erlang.tex
```diff
@@ -991,6 +1021,21 @@
 global variable cannot be used directly because library bindings are
 immutable.
 
+% ----------------------------------------------------------------------------
+\defineentry{with-interrupts-disabled-for-I/O}
```
burgerrg: Should we call it `with-interrupts-disabled-for-io` to be consistent with `wait-for-io` and `complete-io`?

src/swish/erlang.ss
```diff
@@ -208,41 +210,45 @@
...
-
-  (define (pcb-interrupt? p)
-    (fxlogbit? 3 (pcb-flags p)))
-
-  (define (pcb-interrupt?-set! p x)
-    (pcb-flags-set! p
-      (if x
-          (fxlogbit1 3 (pcb-flags p))
-          (fxlogbit0 3 (pcb-flags p)))))
+  (define-syntax (define-pcb-flags x)
```
burgerrg, line 213: Very nice!

src/swish/erlang.ss
```diff
@@ -765,9 +770,70 @@
...
+  ;; its quantum or call receive, so we avoid starving the trap-exit handler by
+  ;; checking again just before an I/O procedure re-enables interrupts.
+  (define-syntax with-interrupts-disabled-for-I/O
+    (syntax-rules ()
+      [(_ e1 e2 ...)
+       (dynamic-wind
+         #%disable-interrupts
+         (lambda () e1 e2 ...)
+         (lambda ()
+           (@check-and-enable-interrupts)))]))
```
burgerrg: Use a single line?

src/swish/erlang.ss
```diff
@@ -765,9 +770,70 @@
...
+  (define (@handle-trap-exit-help)
+    (let ([handle-exit (process-trap-exit)])
+      (cond
+       [(and (procedure? handle-exit) (@get-EXIT-msg)) =>
+        (lambda (msg)
+          (define prior-src (pcb-src self))
+          (define armed? #f)
+          (define content (msg-contents msg))
+          (@remove-msg msg)
+          (enable-interrupts)
```
burgerrg, line 804: Will it cause trouble if the process is interrupted between lines 811 and 812? Or between lines 815 and 816? If so, you could wrap a no-interrupts around them. I think you don't want to move the enable/disable code into the dynamic-wind procedures so that, if the trap-exit handler raises an exception, we don't turn interrupts back on.

doc/swish/erlang.tex
```diff
@@ -973,11 +973,41 @@
...
+\begin{procedure}
+  \code{(raise-on-exit \var{msg})}
+\end{procedure}
+\returns{} see below
+
+The \code{raise-on-exit} procedure expects an exit message
+matching \code{`(EXIT \var{pid} \var{reason})}.
+If \var{reason} is \code{normal}, \code{raise-on-exit} returns
+an unspecified value.
+Otherwise, it raise an exception involving \var{reason}.
```
burgerrg: raise => raises

doc/swish/erlang.tex
owaddell: Fixed, thanks!

doc/swish/erlang.tex
owaddell: Good idea, fixed!

src/swish/erlang.ss
owaddell: :+1:

src/swish/erlang.ss
owaddell: Yep, fixed!

doc/swish/erlang.tex
owaddell: Thanks for catching this, fixed!

src/swish/erlang.ss
owaddell: Hmm, I think we rely on record field accessors being inlined elsewhere in erlang.ss. That could be a problem when we compile with cp0 disabled for profiling.

Maybe we should wrap the calls to `pcb` setters in a `no-interrupts`. The code here isn't performance-critical, so there's no harm, but I think the other places were more frequently executed.

FWIW, here's what the `dynamic-wind` looks like after `expand/optimize`. It looks like the inline code for `#2%fxlogbit0` won't take an interrupt either, though it can call `$oops`.

``` scheme
(#2%dynamic-wind
  (lambda ()
    (if armed?.1
        (#2%errorf #f "attempt to reenter trap-exit handler ~s" handle-exit.2)
        (#2%void))
    (set! armed?.1 #t)
    (let ([g1.3 (#3%$top-level-value 'self.lgnnu3lheosakvgyylzmvq5uw-0)])
      (if (#3%$sealed-record? g1.3 '#<record type pcb>)
          (#2%void)
          (#3%$record-oops 'pcb-src-set! g1.3 '#<record type pcb>))
      (#3%$object-set! 'scheme-object g1.3 129 #f))
    (let ([p.4 (#3%$top-level-value 'self.lgnnu3lheosakvgyylzmvq5uw-0)])
      (let ([g0.5 (#2%fxlogbit1 5
                   (begin
                     (if (#3%$sealed-record? p.4 '#<record type pcb>)
                         (#2%void)
                         (#3%$record-oops 'pcb-flags p.4 '#<record type pcb>))
                     (#3%$object-ref 'scheme-object p.4 105)))])
        (if (#3%$sealed-record? p.4 '#<record type pcb>)
            (#2%void)
            (#3%$record-oops 'pcb-flags-set! p.4 '#<record type pcb>))
        (#3%$object-set! 'scheme-object p.4 105 g0.5))))
  (lambda () (handle-exit.2 content))
  (lambda ()
    (let ([g1.6 (#3%$top-level-value 'self.lgnnu3lheosakvgyylzmvq5uw-0)]
          [g2.7 prior-src.0])
      (if (#3%$sealed-record? g1.6 '#<record type pcb>)
          (#2%void)
          (#3%$record-oops 'pcb-src-set! g1.6 '#<record type pcb>))
      (#3%$object-set! 'scheme-object g1.6 129 g2.7))
    (let ([p.8 (#3%$top-level-value 'self.lgnnu3lheosakvgyylzmvq5uw-0)])
      (let ([g0.9 (#2%fxlogbit0 5
                   (begin
                     (if (#3%$sealed-record? p.8 '#<record type pcb>)
                         (#2%void)
                         (#3%$record-oops 'pcb-flags p.8 '#<record type pcb>))
                     (#3%$object-ref 'scheme-object p.8 105)))])
        (if (#3%$sealed-record? p.8 '#<record type pcb>)
            (#2%void)
            (#3%$record-oops 'pcb-flags-set! p.8 '#<record type pcb>))
        (#3%$object-set! 'scheme-object p.8 105 g0.9)))))
```

owaddell:
I also rewrote the console to fix an issue where the user could evaluate a `(recieve [,x x])` expression and happen to get the `#<cmsg>` record used in my original PR attempt.

I finally added some `expect` scripts to help test the console.

I started by translating the cases from manual-tests.md and then added a few more. I've seen the `simple.exp` test fail sometimes when it's first trying to get off the ground, e.g., it doesn't find the first "> " prompt. Perhaps the timeout is too low?

In theory, we could abstract the uses of ^D in the scripts and make them work under Windows, but the Cygwin install in my VM is so old that it can't update anymore so I haven't pursued this.

edit: To clarify, I did not wire the `expect` scripts into the automate tests; we still have to run the scripts manually.

src/swish/erlang.ss
burgerrg: Thanks for the expansion. I'm satisfied with the current behavior.

test/ReadMe.md
```diff
@@ -0,0 +1,68 @@
+# Console Tests
+
+The console tests here use [Expect](https://core.tcl.tk/expect/index)
```
burgerrg, line 3: Expect works well! I hadn't tried it before, but I was happy to see that it's pre-installed on macOS.

test/ReadMe.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Writing Tests
+
+Keep the following in mind as you adapt the examples in this directory:
+
+1. Include `source common.tcl` near the top of your script.
+   This file provides the `pass`, `fail`, and `must` commands.
+   These commands kill the `swish` process when exiting so that it
+   does not spin out of control and run afoul of the OOM killer.
+2. Use `set sid [spawn swish]`to spawn `swish`. The `pass`, `fail`, and `must`
```
burgerrg: add a space before `to`

src/swish/io.ss
```diff
@@ -109,13 +109,15 @@
...
-   (chezscheme)
+   (except (chezscheme) with-interrupts-disabled)
    (swish compat)
    (swish erlang)
    (swish meta)
    (swish osi)
    )
 
+  (alias with-interrupts-disabled with-interrupts-disabled-for-io)
+
```
burgerrg: Would it be clearer to rename all instances of with-interrupts-disabled within this file?

src/swish/io.ss
```diff
@@ -546,99 +548,109 @@
...
+       (lambda ()
+         (define (do-read bv start n)
+           ;; We still assume that each call has the same bv, start, and n,
+           ;; which the Chez Scheme transcoded-port appears to do.
+           (read-osi-port osip bv start n -1))
+         (let lp ([x #f] [timeout 'infinity] [requests '()] [msg #f])
+           (let ([x (or x (receive (after timeout 'timeout) [,x x]))])
+             (match x
+               [#(read ,bv ,start ,n)
+                (lp #f 0 (cons x requests)
```
burgerrg, line 593: This builds a list of read requests in reverse order, but the list is not processed in reverse order on line 615. Is this a problem? Since there's only one console process that sends read requests one at a time, this list shouldn't go above a single item?

burgerrg: Or maybe this is used for nested read calls after CTRL-C? Then the requests should be processed LIFO as the code does now.

test/ReadMe.md
owaddell: Well spotted. Fixed.

src/swish/io.ss
owaddell: My concern had been that we might miss a case, so I had taken the more draconian approach. But I agree that it's clearer to make the use of this form explicit. We will likely adapt existing code when we add new I/O primitives and we'll follow the pattern.

I pushed a change that tries to make more deliberate use of `with-interrupts-disabled-for-io` focusing on places where we call `wait-for-io` or call something like `@close-osi-port` that itself calls `wait-for-io`.

src/swish/io.ss
owaddell: Yes, I believe the list of `requests` handles nested reads after CTRL-C.

The original code noted that the `transcoded-port` calls `r!` with the same `bv`, `start`, and `n` each time, so we might get away with maintaining a count of pending requests, but we need to get the initial buffer, etc. anyway, hence the list of requests.

burgerrg (approved):

owaddell:
Thanks for the detailed review, Bob!

PR #165, opened 2024-05-01
Minor change to http:options and path-combine

owaddell:
This pull request
- adds `validate-path` to `http:options` and exposes `http:valid-path?` so a server can supply a `http:url-handler` that interprets paths that would otherwise be invalid, yet still check the extracted paths before hitting the filesystem, and
- modifies `path-combine` to avoid adding extra directory separators while allocating less in some cases.
- [edit] fixes a bug in `http:eval-dynamic-page`

### DIFF
--- a/doc/swish/http.tex
+++ b/doc/swish/http.tex
@@ -303,7 +303,12 @@ used:
 
   \code{file-transform} & \code{(lambda (path) \#f)} & a procedure; a
   cache process calls this procedure with a file path and expects an
-  evaluator procedure or \code{\#f} (see \S\ref{sec:dynamic-pages})
+  evaluator procedure or \code{\#f} (see \S\ref{sec:dynamic-pages}) \\
+
+  \code{validate-path} & \code{http:valid-path?} & a predicate;
+  the server raises an exception if the \code{<request>} \var{path}
+  does not satisfy this predicate. \\
+
 \end{tabular}
 
 \defineentry{http:add-server}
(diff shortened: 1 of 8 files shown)


doc/swish/http.tex
```diff
@@ -303,7 +303,12 @@ \subsection{WebSocket Protocol}\label{sec:websocket-protocol}
 
   \code{file-transform} & \code{(lambda (path) \#f)} & a procedure; a
   cache process calls this procedure with a file path and expects an
-  evaluator procedure or \code{\#f} (see \S\ref{sec:dynamic-pages})
+  evaluator procedure or \code{\#f} (see \S\ref{sec:dynamic-pages}) \\
+
+  \code{validate-path} & \code{http:valid-path?} & a predicate;
+  the server raises an exception if the \code{<request>} \var{path}
+  does not satisfy this prediate. \\
```
burgerrg: prediate => predicate

doc/swish/http.tex
```diff
@@ -682,6 +687,16 @@ \subsection{WebSocket Protocol}\label{sec:websocket-protocol}
 Section~\ref{sec:dynamic-page-constructs}, and calls the \code{eval}
 system procedure.
 
+\defineentry{http:valid-path?}
+\begin{procedure}
+  \code{(http:vavlid-path? \var{path})}
```
burgerrg: vavlid => valid

doc/swish/http.tex
owaddell: Oh dear, thanks for finding this. Fixed.

doc/swish/http.tex
owaddell: Thanks for finding this. Fixed.

burgerrg (approved):

laqrix (approved):
Nice work Oscar!

owaddell:
Thanks, guys.

I noticed a failed macOS test and pushed an update to use `test:receive` in that mat.

owaddell:
Oops. One more wafer-thin commit.

owaddell:
Thanks for reviewing this, gentlemen.

PR #167, opened 2024-07-30
release 2.8.0

owaddell:

* Update to SQLite 3.45.3
* Update example.
* Scale another continuous integration timeout.
* Bump version to 2.8.0 and update change log


payneca:
Need to update the release date in the ChangeLog

burgerrg:
Do you want to upgrade to the latest SQLite release, 3.46.0, from May 23?

burgerrg (approved):

owaddell:
Thanks for looking this over, lads.

I'm deferring the SQLite 3.46.0 upgrade to a future release since my internal testing was against 3.45.3 and since 3.46.0 touches the query planner, which has led to SQLite point releases in the past. To their credit, the 3.46.0 release seems to have held up well, but that also suggests that 3.45.3 is solid.

PR #170, opened 2024-12-11
enhance version information

owaddell-ib:
**Fixes**

* migrate off unsupported macOS 12 Actions runners 

**Proposed changes**

* add Chez Scheme `scheme-pre-release` information to `software-info` if applicable
* remove `SWISH_VERSION` from `configure` and `swish.h`
* use `git describe` to determine the version based on tag
* check that the resulting version identifier is compatible with semver 2.0
* work around some flaky macOS tests


burgerrg (approved):
Thank you!

owaddell:

PR #171, opened 2024-12-12
cope with GitHub's ubuntu-24.04 rollout

owaddell-ib:
Explicitly install some packages we depend on that are no longer present in the ubuntu-latest runner when it happens to be ubuntu-24.04. Do this for the `build` job that caches boot files and for the `mats` jobs that actually run the tests, since those will eventually move beyond 22.04.


laqrix (approved):
Looks good. Thanks Oscar.

owaddell:
Thanks for the quick turn-around, Chris.

PR #173, opened 2025-03-12
lift some work out of pregexp-match-positions-aux

owaddell-ib:
The pregexp-match-positions-aux procedure was burrowing through the entire regexp to construct an alist of backrefs. That can be inefficient since pregrexp-match-positions calls it in a loop. Instead, we construct the list of backref keys at expand time, if possible.

Note: be sure to set (print-graph #t) if you expand uses of the (re ...) macro, write the expanded output to a port, and attempt to read and evaluate the expanded code. (This practice is neither common nor recommended.)

### DIFF
--- a/src/swish/pregexp.ss
+++ b/src/swish/pregexp.ss
@@ -17,6 +17,7 @@
 ;;; - added re macro for compile-time parsing
 ;;; - updated replace functions to use a string output port for efficiency
 ;;; - removed unused sn argument from pregexp-match-positions-aux
+;;; - reduced construction of backrefs list in pregexp-match-positions-aux
 ;;; - lookbehind matching handles newlines and honors non-zero start
 ;;; - removed *pregexp-version*
 ;;; - inlined *pregexp-comment-char*, *pregexp-nul-char-int*,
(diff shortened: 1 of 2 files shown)


src/swish/pregexp.ss
```diff
@@ -658,15 +669,17 @@
...
-                  [(string? pat) (pregexp pat)]
-                  [(pair? pat) pat]
-                  [else (pregexp-error 'pregexp-match-positions
-                          'pattern-must-be-compiled-or-string-regexp
-                          pat)])])
+      (let-values ([(pat backrefs)
+                    (match pat
+                      [`(<pregexp> ,pat ,backrefs) (values pat backrefs)]
+                      [,_ (guard (string? pat)) (pregexp-help pat)]
+                      [,_ (guard (pair? pat)) (pregexp-help-pair pat)]
```
burgerrg, line 676: Nice touch keeping support for the previous output of pregexp here.

burgerrg (approved):
Thanks for the clever improvement!

burgerrg (approved):

PR #174, opened 2025-04-26
fix issue with custom-write of list or JSON object

owaddell:
When a `json:custom-write` procedure calls `wr` in tail position, the list and JSON object cases could return `#f` (i.e., when indent is `#f`). This led to `custom-write` returning `#f` to the internal JSON writer, which then continued processing with the remaining built-in cases. As a result, the input could be rejected even when the custom writer believed it had successfully handled it.


burgerrg (approved):
Nice find & fix!

owaddell:
Thanks for the review, Bob!

PR #175, opened 2025-06-17
ci: remove windows-2019 runner image usage

payneca:
https://github.com/actions/runner-images/issues/12045


burgerrg (approved):
Thanks for keeping the runner list up to date!

PR #176, opened 2025-06-21
add osi_get_available_parallelism

payneca:
I'm not sold on the name, but it matches our style of `osi_get` prefix on `uv_available_parallelism`.


burgerrg (approved):
`osi_get_parallelism` would also be fine with me.

payneca:
Thanks for the quick review Bob. After further consideration and noting other implementations stick with "available parallelism", I decided we'll stick with the full name. We can always add a helper later if we find it unwieldy.

PR #177, opened 2025-06-24
test: make db marshaling tests faster

owaddell-ib:


burgerrg (approved):
Nice work using a single db query!

owaddell-ib:

PR #180, opened 2025-10-23
fix define-tuple is? predicate to handle alias or library rename

owaddell-ib:
**Fixes**

Fix the macro generated by define-tuple so that the `is?` predicate expects the original name despite the use of alias and library renaming, just as `make` was fixed in 32eea8d7c93784fdb6146aeed2960f7ac5b32f26.

**Proposed changes**

* Add regression tests for other exported macros that generate macros that could be susceptible to a similar bug.
* Update the Github test runners and test matrix:
   - remove 32-bit builds for all but i3le.
   - remove v9.6.4 builds for all but a6le; macos-15-intel could not build v9.6.4
   - remove builds that use the (unsupported) threaded Chez Scheme runtime

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,6 +1,8 @@
 ---
 name: Test
 on: [push]
+env:
+  RELEASED: v10.2.0
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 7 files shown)


burgerrg (commented):
We should update README.md to say that we don't support threaded Chez Scheme.

owaddell-ib:
Okay, I tried to clarify the statement in the README.

burgerrg (approved):
Thanks for these updates!

owaddell:
Thanks for the review, Bob!

PR #181, opened 2026-01-27
Profiler enhancements

payneca:
- Enable profiler to load the output of `(profile-dump)`
- Allow profiler to ingest absolute paths
- Eliminate `<profile-data>`
- Fix documentation typo

### DIFF
--- a/doc/swish/app.tex
+++ b/doc/swish/app.tex
@@ -371,7 +371,7 @@ the directory specified by \code{base-dir}.
 
 \defineentry{include-line}
 \begin{syntax}
-  \code{(include-line \var{filenme} \opt{\var{not-found}})}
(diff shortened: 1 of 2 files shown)


doc/swish/app.tex
burgerrg, line 1: Nice catch!

burgerrg (approved):
Thanks for the simplification!

owaddell (approved):
Thanks, Chris. This is a nice addition and cleanup.

Folks may have trouble loading old Swish profile output into the new version, but that's never been guaranteed to work. The long-standing use of `fasl` form means that recompiling with a different version of Scheme could already have affected that hypothetical use case.

payneca:
Last minute update. Changed to include `integer?` and `exact?`.

PR #182, opened 2026-01-29
add common-path-prefix to fix profile report bug

owaddell-ib:
**Fixes**

When eliding common prefixes of file paths, the profile report can incorrectly treat `"/home/bob/project"` as a prefix of `"/home/bob/project-backup"` (string prefix does not imply valid path prefix).

**Proposed changes**

* Add `common-path-prefix` to `(swish io)` that operates as if on path components, but without the extra allocation of `path-parent`, et al.
* Use `common-path-prefix` when generating a profile report.

### DIFF
--- a/doc/swish/erlang.tex
+++ b/doc/swish/erlang.tex
@@ -1810,6 +1810,22 @@ fixnum between 0 and 65535 inclusive or a string, exception
 \code{\#(bad-arg connect-tcp \var{port-spec})} is raised.
 
 % ----------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


burgerrg:
It might be worth noting in the description that it compares the filenames in a case sensitive way even if the underlying filesystem is case insensitive.

burgerrg (approved):
Good catch finding the problem with the profile common-prefix detection code and for this helpful function!

owaddell:
Thanks for the review, Bob!

PR #183, opened 2026-02-20
libuv 1.52.0

burgerrg:
Upgrade libuv from 1.48.0 to 1.52.0.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,7 +2,7 @@
 name: Test
 on: [push]
 env:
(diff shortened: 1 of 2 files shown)


owaddell (approved):
Thanks for the update, Bob! Looks like they've made some useful fixes.

I pushed a proposed tidy-up for your consideration. You'll want to remove or incorporate it before merging.

burgerrg:
Thanks, I incorporated the tidy-up with one small change!

PR #185, opened 2026-03-30
add JSON5 support

payneca:
**Fixes**

* fixed an issue where reading a JSON string immediately after a failure to read a string resulted in incorrect output.

* fixed an issue where reading a JSON float with a large exponent consumed CPU.

**Proposed changes**

* require Chez Scheme 10.3.0; drop 9.6.4

* breaking: change representation of JSON null from the character `#\nul` to the symbol `null`

* breaking: change JSON read and write functions to use options
  This introduces new `json:read-options` and `json:write-options`.
  `custom-inflate` is now the `inflate-object` field of `json:read-options`.
  `custom-write` is now the `custom-write` field of `json:write-options`.

* breaking: move stack->json and json-stack->string to json.ss
  Code that used `(import (swish log-db))` may now need `(import (swish json))` to access these procedures.

* A new `json5?` option allows disabling JSON5 support. By default JSON5 is enabled.

* A new `inflate-symbol` option allows raw symbols in JSON-like input to be internalized using a custom function.

* A new `extended-identifiers?` option allows raw symbols in JSON-like input to include characters not allowed by the JSON5 identifier syntax.

* split JSON documentation into a separate chapter

* export `json:key=?`


burgerrg:
Now that we don't support Chez 9.6, we should be able to shrink src/swish/compat.ss and possibly move the definitions elsewhere.

burgerrg:
In src/swish/software-info.ss, we no longer need to check `(top-level-bound? 'scheme-pre-release)`.

src/swish/erlang.ss
```diff
@@ -1950,6 +1950,7 @@
           (bad-arg 'print-radix x))
         x)))
   (redefine print-record (make-process-parameter #t (lambda (x) (and x #t))))
+  (redefine print-subnormal-precision (make-process-parameter #t (lambda (x) (and x #t))))
```
burgerrg, line 1938: Good find! Should we redefine `print-extended-identifiers` as well?

burgerrg, line 1938: I don't like Chez Scheme's default for `print-subnormal-precision` and am sorry I didn't realize it before it went in. We might change the default to `#f` for swish.

src/swish/erlang.ss
burgerrg: We can eliminate this now and change the uses below to redefine. Look for v9.5.8.

doc/swish/json.tex
```diff
@@ -0,0 +1,535 @@
...
+The \var{options} are defined with \code{(json:write-options
+  (\var{option} \var{value}) \etc)}. The following options may be
+used:
+
+\begin{tabular}{lrp{.5\textwidth}}
+  option & default & description \\ \hline
+
+  \code{custom-write} & \code{\#f} & The \var{custom-write} procedure
+  may intervene to handle lists and hashtables differently or to
+  handle objects that have no direct JSON counterpart.  If
```
burgerrg: Please clarify when `custom-write` is called. My reading of the code indicates that it's called on any object that is not a boolean, 'null, string, fixnum, bignum, or flonum.

doc/swish/json.tex
```diff
@@ -0,0 +1,535 @@
...
+  \code{value} & the result of \code{\fixtilde(format "~s" \var{val})} \\
+\end{tabular}
+
+\defineentry{json-stack->string}
+\begin{procedure}
+  \code{(json-stack->string \opt{\var{op}} \var{x})}
+\end{procedure}
+\returns{} see below
+
+The two argument form of \code{json-stack->string} prints the
```
burgerrg: Add a hypen: two-argument form

doc/swish/json.tex
```diff
@@ -0,0 +1,535 @@
...
+stack represented by JSON object \var{x} to the textual output port \var{op}.
+The single argument form of \code{json-stack->string} prints the stack
```
burgerrg: single-argument

burgerrg (commented):
I've reviewed the documentation and like the changes!

burgerrg:
Please update .github/workflows/test.yml to use actions/checkout@v6.

src/swish/erlang.ss
payneca: We're going to go ahead and inherit Chez Scheme's defaults.

burgerrg:
src/swish/swish-build.ms has comments about 9.5.3 error messages. These can be removed now.

doc/reference.bib
```diff
@@ -83,6 +83,14 @@ @misc{html5
   note={\url{http://www.w3.org/TR/2014/REC-html5-20141028/}}
 }
 
+@misc{json5,
+  author={Aseem Kishore and Jordan Tucker},
+  title = {{The JSON 5 Data Interchange Format}},
```
burgerrg: The web page uses `JSON5` without a space before the `5`.

src/swish/errors.ss
```diff
@@ -74,14 +74,20 @@
...
       [#(invalid-context ,who) (format "Invalid context for ~a." who)]
-      [#(invalid-datum ,x) (format "Invalid datum: ~s." x)]
       [#(invalid-intensity ,x) (format "Invalid intensity: ~s." x)]
-      [#(invalid-number ,x) (format "Invalid number: ~s." x)]
       [#(invalid-owner ,owner) (format "Invalid owner: ~s." owner)]
       [#(invalid-period ,period) (format "Invalid period: ~s." period)]
       [#(invalid-procedure ,proc) (format "Invalid procedure: ~s." proc)]
       [#(invalid-strategy ,x) (format "Invalid strategy: ~s." x)]
       [#(io-error ,name ,who ,errno) (format "I/O error ~d from ~a on ~a: ~a." errno who name (errno->english errno))]
+      [#(json:unexpected ,context ,what ,position ,name)
```
burgerrg, line 83: Could we use a symbol for the context instead of the English strings "object key", "value", "string", etc.? It would be consistent with avoiding strings in English. (It's currently possible for another natural language to match these strings and translate them, but we typically do so with symbols.)

src/swish/software-info.ss
```diff
@@ -126,8 +125,7 @@
     [,cs-version
      (format "~a~@[-pre-release.~a~]"
        (scheme-version)
-       (and (top-level-bound? 'scheme-pre-release)
-            (eval '(scheme-pre-release))))]
+       (scheme-pre-release))]
```
burgerrg: This is now short enough to fit on one line.

src/swish/json.ss
```diff
@@ -510,11 +877,20 @@
...
+     [(bignum? x) (display-string (number->string x) op)]
+     [(flonum? x)
+      (cond
+       [(finite? x)
+        (parameterize ([print-precision #f] [print-subnormal-precision #f])
+          (display-string (number->string x) op))]
+       [(not json5?)
+        (throw `#(json:invalid-datum ,x))]
+       [(eqv? x +inf.0) (display-string "Infinity" op)]
+       [(eqv? x -inf.0) (display-string "-Infinity" op)]
```
burgerrg: Consider using the more efficient `#3%fl=` instead of `eqv?`.

src/swish/json.ss
```diff
@@ -510,11 +877,20 @@
...
+       [else (display-string "NaN" op)])]
      [(and custom-write (custom-write op x indent))]
```
burgerrg: The custom writer can return any non-false value to indicate success. We should return void here to prevent that truthy value from escaping.
`[(and custom-write (custom-write op x indent)) (void)]`

src/swish/json.ss
burgerrg, line 949: This code doesn't handle cyclic or improper lists. I wasn't concerned about this in the past, but now I wonder if we could get leverage from using the built-in `list?` to guard these at the door?

burgerrg:
On line 1386 of erlang.tex, we find:
`\var{max-depth} and stops if recognizes a \code{limit-stack} frame.`

It should say:
`\var{max-depth} and stops if it recognizes a \code{limit-stack} frame.`

src/swish/json.ss
```diff
@@ -23,26 +23,31 @@
...
    json:pretty
    json:read
+   json:read-options
    json:ref
    json:set!
    json:size
    json:string->object
    json:update!
    json:write
    json:write-object
```
burgerrg, line 46: `json:write-object` calls `eval` because of phasing problems. The `write-string` procedure deals with this via a macro.

Could we instead put everything else into a `(swish json-impl)` library and then define `json:write-object` in `(swish json)` as well as re-export everything from `(swish json-imp)`? Then we'll get the proper phasing without any tricks.

src/swish/json.ss
```diff
@@ -305,17 +563,18 @@
                  (read (open-input-string s))))
```
burgerrg, line 584: We could use `#3%` on the string-ref and fx- primitives since `string-length` will guarantee we have a string. Type recovery automatically used `#3%char=?`.

src/swish/json.ss
burgerrg: Is the value we get from having a custom equivalence procedure more than the efficiency we'd get from using a stock symbol hashtable?

src/swish/json.ss
```diff
@@ -139,74 +170,253 @@
...
-    (declare-unsafe-primitives read-char)
-    (let ([c (read-char ip)])
-      (if (ws? c)
-          (next-non-ws ip)
-          c)))
+  (define (ws? x json5?)
+    (cond
+     [(not json5?)
+      (memv x '(#\x20 #\x09 #\x0A #\x0D))]
+     [(char-whitespace? x)]
```
burgerrg: Unfortunately, `char-whitespace?` returns true for #\x85 (next line, category Cc, not Zs). I think we should let it through, because the specification likely forgot to include it.

burgerrg:
```
(do ([i 1 (+ i 1)]) [(> i #x10ffff)]
  (when (or (< i #xd800) (> i #xdfff))
    (let* ([c (integer->char i)]
           [ws? (char-whitespace? c)]
           [cat (char-general-category c)])
      (if ws?
          (unless (eq? cat 'Zs)
            (printf "Expected whitespace #\\x~x to have category Zs, not ~a.\n" i cat))
          (when (eq? cat 'Zs)
            (printf "Expected non-whitespace #\\x~x not to have category Zs.\n" i))))))
```

src/swish/json.ss
```diff
@@ -139,74 +170,253 @@
...
+    (let ([c (next-char 'string ip)])
+      (cond
+       [(char=? c mark) (get-json-buffer-string op)]
+       [(char=? c #\\)
+        (or (if json5?
+                (string-escape5 ip op)
+                (string-escape ip op))
+            (unexpected-input 'string (read-char ip) ip))
+        (read-string ip op mark json5?)]
+       [(char<=? c #\x1F)
```
burgerrg: I think JSON5 allows control characters to be unescaped.

burgerrg:
A micro-benchmark on my laptop indicates that performing a generic hashtable-ref costs ~20 ns, whereas a symbol-hashtable-ref costs ~10 ns. Switching (swish json) to use symbol hash tables roughly doubles the speed of lookups. We go from 30 ns/call to 15 ns/call.

```
(define json-object (json:make-object [a #t]))

(define (read-json)
  (json:ref json-object 'a #f))

(define (go f)
  (let lp ([n 1000000])
    (unless (#3%fx<= n 0)
      (f)
      (lp (#3%fx- n 1)))))

> (time (go read-json))
(time (go read-json))
    no collections
    0.030847666s elapsed cpu time
    0.031175000s elapsed real time
    8208 bytes allocated
> (exit)
% git stash apply # To use symbol-hashtable
[...]
> (time (go read-json))
(time (go read-json))
    no collections
    0.015209624s elapsed cpu time
    0.015393000s elapsed real time
    3840 bytes allocated
```
P.S. The allocation appears to occur in other Swish processes, not the benchmark.

src/swish/json.ss
burgerrg: Using a generic hashtable roughly doubles the time required to look up a key. See below for the benchmark I ran to measure this.

burgerrg:
I think all the uses of `zero?` in json.ss are used on exact numbers, so it would be more efficient to use `(eqv? x 0)`.

src/swish/json.ss
burgerrg: Doing `list?` doesn't check for cycles in the car, so let's instead state in the documentation that we don't handle cyclic data and throw an exception when we get an improper list.

src/swish/json.ss
burgerrg: We could do this, but calling eval in these two situations is simple and quick enough not to bother.

burgerrg (approved):
Thanks for this detailed update!

payneca:
Thanks for the review Bob!

PR #187, opened 2026-06-24
Check ,key in json:extend-object

owaddell-ib:
This fixes an issue Chris discovered where `(json:make-object [,"foo" 3])` failed to detect that `"foo"` did not evaluate to a symbol.

The attempt at fixing Windows builds is guesswork. Early CI failures were at the point where we try to build Chez Scheme itself; updating the CI build to use the latest release helped, and we needed to make a few tweaks to adapt to changes in the Windows runner and provide yet more timing flexibility for the overloaded macOS runner.

### DIFF
--- a/src/swish/json.ss
+++ b/src/swish/json.ss
@@ -95,10 +95,15 @@
          ...
          ht)]))
 
+  (define (ensure-key x src)
+    (unless (symbol? x)
+      (throw `#(json:invalid-key ,x ,src)))
+    x)
+
   (define-syntax (parse-key x)
     (syntax-case x (unquote)
       [(_ id form) (identifier? #'id) #'(quote id)]
-      [(_ (unquote e) form) #'e]
+      [(_ (unquote e) form) #`(ensure-key e #,(find-source #'form))]
       [(_ key form) (syntax-error #'form (format "invalid key ~s in" (datum key)))]))
 
   (define (verify-json-object who x)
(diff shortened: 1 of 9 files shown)


src/swish/json.ss
```diff
@@ -95,11 +95,20 @@
...
+  (define (ensure-key x form)
+    (unless (symbol? x) (reject-key x form))
+    x)
+
   (define-syntax (parse-key x)
     (syntax-case x (unquote)
       [(_ id form) (identifier? #'id) #'(quote id)]
-      [(_ (unquote e) form) #'e]
-      [(_ key form) (syntax-error #'form (format "invalid key ~s in" (datum key)))]))
+      [(_ (unquote e) form) #'(ensure-key e #'form)]
```
burgerrg: This quotes the syntax object at run time. Would we prefer to use `find-source` to quote just the source information and then raise a regular run-time error with it? It's a bit unusual to get a syntax-error at run time.

owaddell-ib: Yes. Using `syntax-error` was a tad sleazy. While I doubted anyone would get terribly upset about the exception type, retaining source information in the form `find-source` provides is better because it's less overhead than retaining a syntax object. Just contributing to the training data so LLMs get better at Scheme code.

burgerrg (approved):
Thanks for this update!

owaddell-ib:
Thanks for the review, Bob!

